PR #2, opened 2016-07-24
validate incoming data

Dieterbe:
* we need to reject faulty data
* clients need feedback when their data is wrong


woodsaj:
this needs to be fixed before merging.  It is going to have a significant performance hit due to the passed metrics payloads being iterated over twice when the sender isAdmin(), ie the majority of our metrics.

PR #4, opened 2016-07-26
fix: don't include past msg data into current data

Dieterbe:
but this way we can reuse the existing backing array

### DIFF
--- a/metric_publish/publish.go
+++ b/metric_publish/publish.go
@@ -85,7 +85,7 @@ func Publish(metrics []*schema.MetricData) error {
 	pre := time.Now()
 
 	for i, metric := range metrics {
-		data, err = metric.MarshalMsg(data[:])
+		data, err = metric.MarshalMsg(data[len(data):])
 		if err != nil {
 			return err
 		}


metric_publish/publish.go
```diff
@@ -85,7 +85,7 @@ func Publish(metrics []*schema.MetricData) error {
 	pre := time.Now()
 
 	for i, metric := range metrics {
-		data, err = metric.MarshalMsg(data[:])
+		data, err = metric.MarshalMsg(data[len(data):])
```
woodsaj: I dont think this will provide any performance gain.  Instead of creating many slices that start of small and grow as metric.MarshalMsg() appends bytes to them, you are now just using 1 slice that starts small and continually has bytes appended to it.  There is no change in the number of bytes allocated, as there is really no difference between appending  1 byte to 10 different slices and appending a 1 byte 10 times to 1 slice.

Dieterbe: i'm not sure what the default size is and whether extra allocations are needed at all.
if the initial array is adequality sized we may not need any allocation.
an allocation is pretty expensive though, irrespective of how many bytes are allocated. and i believe the go runtime always allocates about 2x current len when it runs out of space, so the larger the array already is, the fewer allocations are needed.

anyway, all this is pretty speculative. let's close this for now and if/when it becomes a real problem we can do the optimision work with detailed measurements.

PR #17, opened 2016-12-20
accpet raintank/schema.V0 metrics

woodsaj:
Some customers are still sending V0 metrics. When decoded into
a schema.V1 metric the metric.Mtype will be empty and the metric.Metric
may also be empty. These fields need to be populated to pass validation.

### DIFF
--- a/api/metrics.go
+++ b/api/metrics.go
@@ -104,6 +104,13 @@ func metricsBinary(ctx *Context, compressed bool) {
 
 		if ctx.IsAdmin {
 			for _, m := range metricData.Metrics {
+				if m.Metric == "" {
+					m.Metric = m.Name
+				}
+				if m.Mtype == "" {
+					m.Mtype = "gauge"
+				}
+
 				if err := m.Validate(); err != nil {
 					ctx.JSON(400, err.Error())
 					return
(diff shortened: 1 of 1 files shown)


api/metrics.go
```diff
@@ -104,6 +104,13 @@ func metricsBinary(ctx *Context, compressed bool) {
 
 		if ctx.IsAdmin {
 			for _, m := range metricData.Metrics {
+				if m.Metric == "" {
+					m.Metric = m.Name
+				}
+				if m.Mtype == "" {
+					m.Mtype = "gauge"
```
woodsaj: Yes but there is nothing that can be done about it.  Either we set the type to a static value or we reject the message.

Dieterbe:
this takes some shortcuts that are technically incorrect but are acceptable as long as we know for a fact people are only sending data using carbon-relay-ng (which uses gauges, and sets Metric and Name property to the same thing). If they send data using anything else, this code does the wrong thing.

I would rather ask the customers to start using v1 TBH

woodsaj:
100% agree users should just update and send v1 metrics.  But until that happens we run the risk of their tsdb-gw being inadvertently upgraded and their metrics being rejected

api/metrics.go
Dieterbe: unless there was a way to somehow identify whether an incoming request is v0 or v1.
but afaik the http request looks the same, and the msgp []byte has no information in it to distinguish.
we could also decode the msp data into a struct that has an Mtype as well as a TargetType field, and then copy that into a MetricData based on what it has.  But that might be taking it too far.

woodsaj: i think that is unnecessary.  As far as i am concerned, this is just a short term work around. Once we have confirmed everyone has upgraded their carbon-relay-ng we can remove it.

PR #18, opened 2016-12-20
the latest version of sarama requires config.Producer.Return.Success to be set

woodsaj:
see https://github.com/Shopify/sarama/pull/790


woodsaj:
This needs to be merged promptly as the tsdb-gw latest tsdb-gw packages are broken and dont run.

PR #22, opened 2017-01-02
replace epoch_millis with empty string to avoid error, add basic tests

DanCech:


elasticsearch/elasticsearch.go
```diff
@@ -112,11 +112,12 @@ type esSearch struct {
...
 
 type esQuery struct {
-	Filtered esFiltered `json:"filtered"`
+	Bool     interface{} `json:"bool,omitempty"`
+	Filtered esFiltered  `json:"filtered"`
 }
 
 type esFiltered struct {
-	Query  interface{} `json:"query"`
+	Query  interface{} `json:"query,omitempty"`
```
DanCech: These changes were necessary to support the format seen in queries from Grafana, where the query element contains the search terms directly in `query.bool.must` rather than under `query.filtered.filter.bool.must`

DanCech:
Again here I'm unsure about the best way to test this, so we need to do that and verify that it actually works.

In particular, given what I'm seeing with the query format that Grafana sends, I'm not sure how the existing code could work.

DanCech:
That seems to work much better, and it's completely agnostic about the structure of the query provided by the user, since it just wraps it:
```
{
  "bool": {
    "must": [
      <user query>
    ],
    "filter": {
      "term": {
        "org_id": <org id>
      }
    }
  }
}
```

PR #25, opened 2017-01-30
proxy delete endpoint to metrictank

### LINKED ISSUE #21: forward delete api endpoint to metrictank
so that customers can delete their own series
everybody saves time, both customers and us -> win-win

replay:
directly proxy requests on `/metrics/delete` to the metrictank
endpoint, bypassing graphite-api.

fixes #21 

example:
```
mst@ubuntu:~$ curl  -H 'X-Org-Id: 1' 'http://localhost/metrics/delete' --form 'query=some.id.of.a.metric.1'
{"deletedDefs":1}
mst@ubuntu:~$ curl  -H 'X-Org-Id: 1' 'http://localhost/metrics/delete' --form 'query=some.id.of.a.metric.1'
{"deletedDefs":0}
```

### DIFF
--- a/metrictank/metrictank.go
+++ b/metrictank/metrictank.go
@@ -1,0 +1,34 @@
+package metrictank
+
+import (
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"strconv"
+
+	"github.com/raintank/tsdb-gw/util"
+)
+
+var (
+	MetrictankUrl *url.URL
+)
+
+func Init(metrictankUrl string) error {
+	var err error
+	MetrictankUrl, err = url.Parse(metrictankUrl)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func ProxyDelete(orgId int64) *httputil.ReverseProxy {
+	director := func(req *http.Request) {
+		req.URL.Scheme = MetrictankUrl.Scheme
+		req.URL.Host = MetrictankUrl.Host
+		req.URL.Path = util.JoinUrlFragments(MetrictankUrl.Path, "/metrics/delete")
+		req.Header.Del("X-Org-Id")
+		req.Header.Add("X-Org-Id", strconv.FormatInt(orgId, 10))
+	}
+	return &httputil.ReverseProxy{Director: director}
+}
--- a/api/api.go
+++ b/api/api.go
@@ -19,6 +19,7 @@ func InitRoutes(metrics met.Backend, m *macaron.Macaron, adminKey string) {
 	m.Use(RequestStats())
 
 	m.Get("/", index)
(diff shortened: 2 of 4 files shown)


metrictank/metrictank.go
```diff
@@ -0,0 +1,34 @@
...
+	return err
+}
+
+func ProxyDelete(orgId int64) *httputil.ReverseProxy {
+	director := func(req *http.Request) {
+		req.URL.Scheme = MetrictankUrl.Scheme
+		req.URL.Host = MetrictankUrl.Host
+		req.URL.Path = util.JoinUrlFragments(MetrictankUrl.Path, "/metrics/delete")
+		req.Header.Del("X-Org-Id")
+		req.Header.Add("X-Org-Id", strconv.FormatInt(orgId, 10))
```
Dieterbe: so the goal here is to set the proper x-org-id header, overriding anything the user may have passed right? I wonder if it also works if the user sets the header with different casing. 
I thinkit should work, from the comments in the go source code,it does this:
> // CanonicalMIMEHeaderKey returns the canonical format of the
// MIME header key s. The canonicalization converts the first
// letter and any letter following a hyphen to upper case;
// the rest are converted to lowercase. For example, the
// canonical key for "accept-encoding" is "Accept-Encoding".
// MIME header keys are assumed to be ASCII only.
// If s contains a space or invalid header field bytes, it is
// returned without modifications.

but to be 100% sure, could you try it? try passing custom headers like `x-org-id`, `x-org-ID` and `x-Org-id` and confirm nothing makes it past tsdb-gw

woodsaj: HTTP header names are case-insensitive.  https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

replay: @Dieterbe i've tested it to be completely sure. I added `fmt.Printf("Header: %+v", req.Header)` right after the `Del()`/`Add()` on the header. Then I requested like this:

```
mst@ubuntu:~$ curl  -H 'x-org-id: 1' 'http://localhost/metrics/delete' --form 'query=some.id.of.a.metric.4'
{"deletedDefs":0}
```
That's the `tsdb-gw` output:
```
Header: map[X-Org-Id:[1] User-Agent:[curl/7.47.0] Accept:[*/*] Content-Length:[161] Expect:[100-continue] Content-Type:[multipart/form-data; boundary=------------------------6f76cc7fd6438646]][Macaron] Completed /metrics/delete 200 OK in 1.648487ms
```

Dieterbe:
thanks @replay , this is just what we needed I think. once we roll this out, our customers will be a bit happier about not needing us anymore for deletes :)

PR #26, opened 2017-01-31
performance improvements

woodsaj:
- clean up config settings.  Moved the configs for event_publisher
  and metric_publisher to their respective packages.
- add tunables for kafka flush duration and max messages.


woodsaj:
This commit changes config variable names, so i think we should also change the version from 0.5.1 to 0.6.0

We then also need to carefully deploy it with the updated config to all environments

PR #28, opened 2017-02-23
Fixes and timeouts

Dieterbe:


graphite/graphite.go
```diff
@@ -1,28 +1,56 @@
...
+	}
+
+	if graphiteTimeout == 0 {
+		return errors.New("graphite-timeout must be a valid non-zero duration")
+	}
+
+	graphiteTransport = &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   graphiteTimeout,
```
woodsaj: This timeout is only for establishing a tcp connection.  That should happen in less then a second.

woodsaj:
I am not sure setting a 30second connect timeout is going to have any affect.  If graphite cant be reached within a few seconds, then waiting 30seconds more probably isn't going to help.

For hosted-metrics, we run tsdb-gw and graphite in the same POD, so tsdb-gw allways connects to localhost:8080.  But of the the graphite container is killed, usually OOM'ed, tsdb-gw will still continue to receive connections from users and try to connect to graphite over localhost.

We can solve this in 2 ways
1) add a /healthz endpoint that to tsdb-gw that checks reachability of all proxy targets, and fails if any cant be reached.  This would remove the tsdb from service.
2) split graphite into its own POD.

2 is probably going to yield better results.

graphite/graphite.go
Dieterbe: oops. this was supposed to control the timeout of waiting for a response from the upstream server.
and the reason is i'm seeing these in the tsdb-gw logs for one of our customers who does large queries:
```
2017/02/23 13:00:24 http: proxy error: net/http: request canceled
[Macaron] Completed /graphite/render 502 Bad Gateway in 29.671191282s
```

Dieterbe:
i changed it to clarify it's the dial timeout, and changed it from 60s to 30s (== default in Go). I actually don't think I would want to wait 60s for a dial to fail. in fact, once you implement your change number 2, we should probably shorten it to a couple seconds and make it retry. the loadbalancer then takes care of the rest.

PR #43, opened 2017-07-19
add usage reporting

woodsaj:
- created separate tsdb-usage binary that accepts and aggregates
  usage data from 1 or more tsdb-gw servers.  Each org has their
  own usage metrics and these metrics are fed into metrictank
  via kafka.
- the metrics collected are
  - hosted-metrics.usage.active_series : unique number of series seen in previous 6hours
  - hosted-metrics.usage.datapoints_per_minute : rate of metrics received
  - hosted-metrics.usage.api.request.`<path>`.status.`<status>` : api request counts by response status.

### DIFF
--- a/cmd/tsdb-usage/main.go
+++ b/cmd/tsdb-usage/main.go
@@ -1,0 +1,135 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"flag"
+	"io"
+	"net"
+	"os"
+	"os/signal"
+	"strconv"
+	"sync"
+	"syscall"
+
+	"github.com/golang/glog"
+	"github.com/grafana/globalconf"
+	"github.com/raintank/tsdb-gw/metric_publish"
+	"github.com/raintank/tsdb-gw/usage"
+)
+
+var (
+	broker   = flag.String("kafka-tcp-addr", "localhost:9092", "kafka tcp address for metrics")
+	addr     = flag.String("addr", "localhost:2004", "listen address")
+	confFile = flag.String("config", "/etc/raintank/tsdb.ini", "configuration file path")
+)
+
+func main() {
+	flag.Parse()
+	// Only try and parse the conf file if it exists
+	path := ""
+	if _, err := os.Stat(*confFile); err == nil {
+		path = *confFile
+	}
+	conf, err := globalconf.NewWithOptions(&globalconf.Options{
+		Filename:  path,
+		EnvPrefix: "TU_",
+	})
+	if err != nil {
+		glog.Fatalf("error with configuration file: %s", err)
+	}
+	conf.ParseAll()
+
+	glog.Info("tsdb-usage starting up")
+
+	// initialize our publisher that sends metrics to Kafka
+	metric_publish.Init(*broker)
+
+	interrupt := make(chan os.Signal, 1)
+	signal.Notify(interrupt, os.Interrupt, syscall.SIGTERM)
+
+	// define our own listener
+	l, err := net.Listen("tcp", *addr)
+	if err != nil {
+		glog.Fatal(4, err.Error())
+	}
+	glog.Infof("tsdb-usage listening on %s", *addr)
+
+	// cleanly shutdown when we are asked to
+	go handleShutdown(interrupt, l)
+	Init()
+	StatsRun()
+	listen(l)
+}
+
+func handleShutdown(interrupt chan os.Signal, l net.Listener) {
+	<-interrupt
+	glog.Info("shutdown started.")
+	l.Close()
+}
+
+func listen(l net.Listener) {
+	shutdown := make(chan struct{})
+	var wg sync.WaitGroup
+	for {
+		// Listen for an incoming connection.
+		conn, err := l.Accept()
+		if err != nil {
+			glog.Infof("listener error. %v", err)
+			break
+		}
+		// Handle connections in a new goroutine.
+		wg.Add(1)
+		go handleRequest(conn, shutdown, &wg)
+	}
+	close(shutdown)
+	wg.Wait()
+	glog.Info("listener has shutdown.")
+	return
+}
+
+func handleRequest(conn net.Conn, shutdown chan struct{}, wg *sync.WaitGroup) {
+	defer func() {
+		glog.Infof("connection from %s ended", conn.RemoteAddr().String())
+		wg.Done()
+	}()
+
+	// if shutdown is received, close the connection.
+	go func() {
+		<-shutdown
+		conn.Close()
+	}()
+
+	glog.Infof("new connection from %s", conn.RemoteAddr().String())
+	r := bufio.NewReaderSize(conn, 4096)
+	var org int64
+	for {
+		buf, _, err := r.ReadLine()
+		if err != nil {
+			if err == io.EOF {
+				glog.Infof("EOF from %s", conn.RemoteAddr().String())
+			} else {
+				glog.Errorf("connect error from %s. %v", conn.RemoteAddr().String(), err)
+			}
+			conn.Close()
+			return
+		}
+		eType := usage.EventType(buf[0])
+		if !(eType == usage.DataPointReceived || eType == usage.ApiRequest) {
+			glog.V(4).Infof("unknown event type %c", eType)
+			continue
+		}
+		// the data payload is in the form <org>.<metric>
+		data := bytes.SplitN(buf[1:], []byte{'.'}, 2)
+		org, err = strconv.ParseInt(string(data[0]), 10, 32)
+		if err != nil {
+			glog.Errorf("invalid data. could not parse org. %v", err)
+			continue
+		}
+		Record(&usage.Event{
+			Org:   int32(org),
+			ID:    string(data[1]),
+			EType: eType,
+		})
+	}
+}
--- a/cmd/tsdb-usage/stats.go
+++ b/cmd/tsdb-usage/stats.go
@@ -1,0 +1,216 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"hash/fnv"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/raintank/tsdb-gw/metric_publish"
+	"github.com/raintank/tsdb-gw/usage"
+	"gopkg.in/raintank/schema.v1"
+)
+
+var shardCount = 4
+var bufferSize = 100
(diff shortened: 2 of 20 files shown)


cmd/tsdb-usage/stats.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+var SHARD_COUNT = 4
+var BUFFER_SIZE = 100
+
+var eventsChan chan *usage.Event
+var shards []chan *usage.Event
+
+func init() {
+	flag.IntVar(&SHARD_COUNT, "shard-count", 4, "Number of shards. Should equal number of CPU cores")
```
Dieterbe: can use https://golang.org/pkg/runtime/#GOMAXPROCS

usage/usage.go
```diff
@@ -0,0 +1,172 @@
...
+		case id := <-t.In:
+			t.buf = append(t.buf, id)
+		case <-ticker.C:
+			// non-blocking write to output buffer.
+			// if we succesfully add the current input buffer
+			// to the output chan then we re-initialize the input
+			// buffer to an new empty slice.  If the output chan is
+			// blocked, then we just move on and will retry next flush.
+			select {
+			case t.out <- output{buf: t.buf, ts: time.Now()}:
```
Dieterbe: can just use the ts from ticker.C

usage/usage_test.go
```diff
@@ -0,0 +1,71 @@
...
+	LogRequest(1, fmt.Sprintf("api.request.%s.status.%d", "render", 200))
+
+	LogRequest(1, fmt.Sprintf("api.request.%s.status.%d", "render", 200))
+	for _, s := range series {
+		LogDataPoint(s.Id)
+	}
+
+	LogRequest(1, fmt.Sprintf("api.request.%s.status.%d", "render", 200))
+	Stop()
+}
```
Dieterbe: benchmarks must execute an operation `b.N` times to determine a stable per-operation time.

cmd/tsdb-usage/main.go
```diff
@@ -0,0 +1,135 @@
...
+		// the data payload is in the form <org>.<metric>
+		data := bytes.SplitN(buf[1:], []byte{'.'}, 2)
+		org, err = strconv.ParseInt(string(data[0]), 10, 32)
+		if err != nil {
+			glog.Errorf("invalid data. could not parse org. %v", err)
+			continue
+		}
+		Record(&usage.Event{
+			Org:   int32(org),
+			ID:    string(data[1]),
```
Dieterbe: `data[1:]` I think?

cmd/tsdb-usage/main.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+			conn.Close()
+			return
+		}
+		eType := usage.EventType(buf[0])
+		if !(eType == usage.DataPointReceived || eType == usage.ApiRequest) {
+			glog.V(4).Infof("unknown event type %c", eType)
+			continue
+		}
+		// the data payload is in the form <org>.<metric>
```
Dieterbe: only when it's DataPointsReceived. when it's ApiRequest it needs to be handled differently, no?

Dieterbe: or `<org>.<request-path>`

cmd/tsdb-usage/stats.go
woodsaj: GOMAXPROCS will only limit the number of threads used.  The number of shards is independent of the number of configured or available cores.  The comment is just a guide

cmd/tsdb-usage/main.go
woodsaj: all payloads are in the form <org>.<metric>
for request metrics, <metric> is something like `api.request.<path>.status.<status>`  for dataPoints it is <org>.<metricId>

usage/usage_test.go
woodsaj: fixed

cmd/tsdb-usage/stats.go
```diff
@@ -0,0 +1,209 @@
...
+	ActiveSeries  map[string]uint32
+	RequestCounts map[string]int64
+	DataPoints    int64
+}
+
+type ShardStats struct {
+	Orgs map[int32]*OrgStats
+}
+
+type TsdbOrgStats struct {
```
Dieterbe: some comments here about what these structs are for, specifically why the TsdbStats differ from the OrgStats/Shardstats, would go a long way in making this more understandable.

Dieterbe:
> hosted-metrics.usage.active_series : unique number of series seen in previous 6hours

is there a particular reason why we're switching to 6h ? is that how we bill now?

my main concern with this implementation is the needed network traffic. it's like the "MT statsd problem" : a data line for every ingested point.
With a 1Gbps network, considering every "metric ingest message" is about 100B, that means we saturate the entire link when we do 1M metrics/s; individual tsdbgw servers can't do such volumes due to various inefficiencies (e.g. sending data from client to tsdbgw and from tsdbgw to kafka currently takes many more times the bandwith than the stats messages, so those will saturate the network first; and also they don't have enough cpu power I think), but when we have tens of tsdbgw's being sharded across, and reporting stats to a central usage server, that could become a network problem.

 Also, memory usage of tsdb-usage could become a problem, though we should be able to hold about 100million messages in "only" 20GB ram, with 10s flush rate that's a 10M/s ingestion rate, so i'm not too worried yet.

Here's some proposals on how to address:

1) snappy/gzip/... compression for data from tsdbgw to tsdb-usage and probably get significant (>10x) savings.

2) since we're measuring active series across large timespans, we can pre-aggregate in tsdbgw. e.g. every hour we could compute a list of all id's and how many times they've been seen in that hour. for 30s metrics that's a 120x reduction, before sending the data to tsdb-usage which then further aggregates.
we then have per-hour stats.
note this requires that when tsdbgw shuts down, it flushes pending buffers. when tsdbgw crashes, more data will be lost than with your current approach.

3) tsdb-usage can use HyperLogLog which generates cardinality estimates (typically with 2% error) without having to keep any id in memory. unfortunately we can't use HLL in tsdbgw before submitting to tsdb-usage because the data processed by other tsdbgw instances may be a partially overlapping keyspace. (it would be fine if it was the same, or a different keyspace)

4) since metrics are nicely sharded into kafka partitions, we could have kafka consumers (potentially even running on the broker hosting the partition they're consuming from), and they could all use HLL for the shard, and we could then add up the HLL results for each shard.

woodsaj:
> is there a particular reason why we're switching to 6h ? is that how we bill now?

we should probably make that configurable.  6h is chosen as that is the default metric-max-stale setting used by MT.  Metrics that have not been for this duration are not considered active anymore and are removed from memory.

> With a 1Gbps network, considering every "metric ingest message" is about 100B, that means we saturate the entire link when we do 1M metrics/s;

a) we have 20GB network interfaces on our servers
b) the metric ingest messages are only ~42bytes.  (1byte for eType, up to 7bytes for org (0-9999999), 1byte for separator "." ,  32bytes for metricId and 1byte for end-of-line.

> Also, memory usage of tsdb-usage could become a problem, though we should be able to hold about 100million messages in "only" 20GB ram

my benchmarks also showed 200Mb of memory per 1million series, with linear scaling.  an ingest rate of 10million/s would also require 20 CPUs.    So with current hardware, our maximum capacity is going to be constrained by CPU, at an ingestion rate of ~20million/second

So though there is room for improvement.  It is not needed right now.

usage/usage.go
```diff
@@ -0,0 +1,172 @@
...
+		case id := <-t.In:
+			t.buf = append(t.buf, id)
+		case ts := <-ticker.C:
+			// non-blocking write to output buffer.
+			// if we succesfully add the current input buffer
+			// to the output chan then we re-initialize the input
+			// buffer to an new empty slice.  If the output chan is
+			// blocked, then we just move on and will retry next flush.
+			select {
+			case t.out <- output{buf: t.buf, ts: ts}:
```
replay: is this a case where we could save some allocs by getting the `output{}` from a `sync.Pool`?

usage/usage.go
```diff
@@ -0,0 +1,172 @@
...
+			continue
+		}
+		// dont spend more then 10seconds trying to send data.
+		deadline := time.Now().Add(time.Second * time.Duration(10))
+
+		var err error
+		if t.conn == nil {
+			connected := t.reconnect(deadline)
+			if !connected {
+				// could not connect after 10seconds.  dropping data.
```
replay: maybe this should have a log statement?

usage/usage.go
```diff
@@ -0,0 +1,172 @@
...
+	var err error
+	for deadline.Sub(time.Now()) > time.Duration(0) {
+		t.conn, err = net.DialTCP("tcp", nil, t.addr)
+		if err == nil {
+			connected = true
+			break
+		}
+		log.Error(3, "failed to connect to tsdb-usage server. %v", err)
+		time.Sleep(time.Second)
+		continue
```
replay: what does the `continue` in the end of the `for` do?

cmd/tsdb-usage/stats.go
```diff
@@ -0,0 +1,216 @@
...
+			switch e.EType {
+			case usage.DataPointReceived:
+				stats.DataPoints++
+				stats.ActiveSeries[e.ID] = uint32(time.Now().Unix())
+			case usage.ApiRequest:
+				stats.RequestCounts[e.ID]++
+			}
+
+		case <-cleanUpTicker.C:
+			old := uint32(time.Now().Add(flushInterval).Unix())
```
replay: this `old` is the limit that defines starting from when a time series is considered too old to be kept track off, right? So shouldn't that be `time.Now() - flushInterval` instead of time.Now() + flushInterval`?

replay: also, it seems like if there is somebody sending sparse data at very cardinality quite some memory could be saved if this would run more often than just once an hour. does the looping over the active series take so long that this can't run more often?

usage/usage.go
woodsaj: nothing. it is redundant.  ill remove it.

cmd/tsdb-usage/stats.go
woodsaj: scanning the full list is expensive and it blocks ingestion.  So we dont want to do it too often.
if users are sending extremely dynamic data, which we have seen before when a user accidentally include a timestamp value in the series name, then we want the metrics to reflect that.   Pruning too frequently might hide these types of issues.

PR #48, opened 2017-08-15
Carbon input

woodsaj:
add carbon input plugin
    
- supports carbon input via TCP and UDP
- expects the first element in the series name to be an auth key.
  This key is then looked up in an auth file to map it to an orgId


woodsaj:
> do we want to keep the old metrics.valid

i could not find any dashboards currently using them.  Typically we only look at the number of metrics published to kafka.

woodsaj:
One thing i would like to do is to change the auth handling and make auth plugins that both the http and carbon inputs can use.
this will allow us to switch between using grafana.com api keys as the prefix for carbon ( i dont love this idea as the keys are quite long) and a local auth file. It also allows those running on-prem to use the auth-file instead of eing tied to authenticating against grafana.com

carbon/auth.go
```diff
@@ -0,0 +1,53 @@
...
+
+	scanner := bufio.NewScanner(bufio.NewReader(file))
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "#") || strings.HasPrefix(line, ";") || len(line) < 1 {
+			continue
+		}
+
+		parts := strings.SplitN(line, ":", 2)
+		if len(parts) < 2 {
```
replay: Shouldn't that return an error or warning or something? Otherwise it will be very hard for a user to figure out that their config doesn't work because they typed a `;` instead of `:` or something like that.

carbon/carbon.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (c *Carbon) flush() {
+	defer c.flushWg.Done()
+	buf := make([]*schema.MetricData, 0)
+
+	ticker := time.NewTicker(flushInterval)
+	for {
+		select {
+		case <-ticker.C:
```
replay: So far, when tsdb-gw went down for a short time, we relied on carbon-relay-ng to buffer the data and resend it once tsdb was back. If we're now buffering in the memory of tsdb-gw, doesn't that mean we'll lose data if it dies? it also means that upgrade procedures for tsdb will get more complicated, because we'll need to "drain" them before upgrading each of them.

carbon/carbon.go
```diff
@@ -0,0 +1,258 @@
...
+	for {
+		// Listen for an incoming connection.
+		conn, err := c.tcp.Accept()
+		if err != nil {
+			log.Info("listener error. %v", err)
+			break
+		}
+		// Handle connections in a new goroutine.
+		wg.Add(1)
+		go c.handleRequest(conn, shutdown, &wg)
```
replay: I think the name `handleRequest` is slightly misleading, because afaict that's not handling a request as in HTTP but a connection which is a whole stream of data coming in.

carbon/carbon.go
```diff
@@ -0,0 +1,258 @@
...
+func (c *Carbon) Stop() {
+	if !Enabled {
+		return
+	}
+	c.tcp.Close()
+	c.udp.Close()
+	c.listenWg.Wait()
+	// all listeners are closed, so we can close the input Buf
+	close(c.buf)
+	c.listenWg.Wait()
```
replay: I don't really understand what the second `c.listenWg.Wait()` is for. probably a comment would help

carbon/carbon.go
replay: I think the draining is what the `c.flushWg` is for. Only works on a clean shutdown though, not on f.e. an OOM

replay: I can't see where it is waiting for `flushWg`. Should line `:124` actually wait for that instead of `listendWg`?

carbon/carbon.go
woodsaj: it should be c.flushWg.Wait() which signals that all data in buffers has been flushed to kafka.

carbon/carbon.go
woodsaj: yes `:124` should be c.flushWg.Wait().

And yes, up to 1second (with default flush interval) could be lost on a hard failure.  There really isn't anything that we can do about this as we need to write to kafka in batches otherwise it will have terrible performance.  But in order to write in batches, we need to read lots of metrics into a buffer first.

This is an acceptable limitation. If users want 0 loss, then they will need to send via HTTP and not via carbon.

woodsaj:
Great feedback @replay 

I have refactored the auth to split it into 2 authPlugins (grafanaCOm and file).  Now both the HTTP api and the carbon input plugins can use either auth plugin.

Can you give it another look.

auth/auth.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func GetAuthPlugin(name string) AuthPlugin {
+	switch name {
+	case "grafana":
+		return NewGrafanaComAuth()
+	case "file":
+		return NewFileAuth()
+	default:
+		log.Fatal(4, "invalid auth plugin specified, %s", name)
```
replay: This should probably panic or something. If it returns `nil` then the process would continue starting and only panic once the first request is calling `.Auth()` on it.

carbon/carbon.go
```diff
@@ -42,30 +44,31 @@ func init() {
 	flag.StringVar(&addr, "carbon-addr", "0.0.0.0:2003", "listen address for carbon input")
 	flag.BoolVar(&Enabled, "carbon-enabled", false, "enable carbon input")
 	flag.StringVar(&schemasConf, "schemas-file", "/etc/storage-schemas.conf", "path to carbon storage-schemas.conf file")
-	flag.StringVar(&authFile, "carbon-auth-file", "/etc/carbon-auth", "path to carbon auth file")
 	flag.DurationVar(&flushInterval, "carbon-flush-interval", time.Second, "maximum time between flushs to kafka")
 	flag.IntVar(&concurrency, "carbon-concurrency", 1, "number of goroutines for handling metrics")
 	flag.IntVar(&bufferSize, "carbon-buffer-size", 100000, "number of metrics to hold in an input buffer. Once this buffer fills metrics will be dropped")
+	flag.BoolVar(&nonBlockingBuffer, "carbon-non-blocking-buffer", false, "dont block trying to write to the input buffer, just drop metrics.")
```
replay: I think that's way better. Rather slow the producer down than just throwing its data away

carbon/carbon.go
replay: I think there are things we could do about this, but it would probably not be worth it because it's going to make things more complicated for a very minor benefit.
And anyway it's a good reason to convince people to use http, which should still be recommended.

replay:
Looks good

PR #49, opened 2017-08-24
Jaeger tracing

replay:
Adds open tracing


replay:
This would be ready to review/merge I think @Dieterbe

api/middleware.go
```diff
@@ -176,3 +194,40 @@ func pathSlug(p string) string {
...
+		if status >= 200 && status < 300 {
+			span.SetTag("http.size", rw.Size())
+		}
+		if status >= 400 {
+			span.LogFields(otLog.Error(errors.New(string(rw.errBody))))
+			if status >= 500 {
+				ext.Error.Set(span, true)
+			}
+		}
+		// TODO: else write error msg?
```
Dieterbe: stale todo

graphite/graphite.go
```diff
@@ -34,11 +36,20 @@ func NewProxyRetrytransport() *proxyRetryTransport {
 }
 
 func (t *proxyRetryTransport) RoundTrip(outreq *http.Request) (*http.Response, error) {
+	span := opentracing.SpanFromContext(outreq.Context())
+	span = span.Tracer().StartSpan("graphite_rt", opentracing.ChildOf(span.Context()))
```
Dieterbe: interesting. did not know about `span.Tracer()`

Dieterbe: I think `graphite_rt` is vague. also generally I think we should name spans after functions, so that they're easier to comprehend and trace back to code.in this case this span could be named `graphite.RoundTrip` or better IMHO `graphiteProxy.RoundTrip`

graphite/graphite.go
```diff
@@ -52,7 +63,9 @@ func (t *proxyRetryTransport) RoundTrip(outreq *http.Request) (*http.Response, e
 		if hasBody {
 			outreq.Body = ioutil.NopCloser(bytes.NewBuffer(body))
 		}
+		attempt_span := span.Tracer().StartSpan(fmt.Sprintf("graphite_rt_attempt%d", attempts), opentracing.ChildOf(span.Context()))
```
Dieterbe: this looks a bit wrong. child spans should -I think- not repeat the name of their parent span (they'll be displayed as a tree anyway, so the name of the parent will be right there)
but more interestingly, the span name that was encoded into the http headers does not have attempt information, so if graphite (which typically actually is MT) adds spans into the tree, they will be added to the parent one, not to the proper attempt one and it'll look weird/confusing.

PR #54, opened 2017-09-18
SetReadDeadline on connections to tsdb-gw. fixes #53

### LINKED ISSUE #53: tsdb-usage: out of control amount of tcp sockets

woodsaj:
because we never try and write to connections, if we dont set a
readDeadline we will never notice that a connection is dead and
so the connection will never be closed.

### DIFF
--- a/cmd/tsdb-usage/main.go
+++ b/cmd/tsdb-usage/main.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"sync"
 	"syscall"
+	"time"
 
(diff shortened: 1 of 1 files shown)


Dieterbe:
don't we need to call `Read()` and see if we get EOF back?

woodsaj:
we wouldn't be able to receive any data if we didn't call read()

https://github.com/raintank/tsdb-gw/blob/b36553bf481e2823630403160e8492802989d082/cmd/tsdb-usage/main.go#L109

PR #55, opened 2017-09-19
don't print the spans, it's pointless and wasteful

Dieterbe:


util/tracer.go
```diff
@@ -11,16 +11,15 @@ import (
 )
 
 func GetTracer(enabled bool, addr string) (opentracing.Tracer, io.Closer, error) {
-	// Sample configuration for testing. Use constant sampling to sample every trace
-	// and enable LogSpan to log every span via configured Logger.
+	//  We use constant sampling to sample every trace, until we need better
```
replay: is that sentence complete :) ?

Dieterbe: i see how it's not phrased perfectly, but imho yes :)

PR #58, opened 2017-11-14
add carbon tagging functionality

jtlisi:


carbon/schemas.go
```diff
@@ -30,7 +30,19 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
 		return nil, fmt.Errorf(errFmt3Fields, msg)
 	}
 
-	name := elements[0]
+	metric := strings.SplitN(elements[0], ";", 2)
+
+	tags := []string{}
```
Dieterbe: can use `var tags []string` instead. the benefit is it won't allocate until needed.

carbon/schemas.go
```diff
@@ -42,19 +54,25 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
 		return nil, fmt.Errorf(errFmt, msg, err)
 	}
 
-	_, s := schemas.Match(name, 0)
+	_, s := schemas.Match(metric[0], 0)
 	md := metricPool.Get()
 	*md = schema.MetricData{
-		Name:     name,
-		Metric:   name,
+		Name:     metric[0],
```
Dieterbe: this is not very clear. why not put a `name = metric[0]` near where it used to say `name := elements[0]` ?

carbon/schemas.go
```diff
@@ -42,19 +54,25 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
...
+		Tags:     tags,
 		OrgId:    orgId,
 	}
 	md.SetId()
+
+	err = md.Validate()
+	if err != nil {
+		return nil, err
+
+	}
```
Dieterbe: why is this needed? we should never construct an invalid MD. if we got that far, we should have invalidated incoming data before that.

carbon/schemas.go
```diff
@@ -30,7 +30,19 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
 		return nil, fmt.Errorf(errFmt3Fields, msg)
 	}
 
-	name := elements[0]
+	metric := strings.SplitN(elements[0], ";", 2)
```
Dieterbe: I would suggest do here `metric := strings.Split(elements[0], ";") that way you can later set `tags = metric[1:]` instead of having to split again.

jtlisi: :+1:

jtlisi:
@Dieterbe This should be better

carbon/schemas_test.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		{
+			name: "no key tag metric",
+			args: args{
+				buf:     []byte("test.metric.value;=1 10 10"),
+				schemas: &schemas,
+				orgID:   1,
+			},
+			wantName: "test.metric.value",
+			wantTags: []string{},
```
Dieterbe: you could just use `nil` instead, or even better, don't specify the field, as the default value is already nil.
in fact, same thing for wantName. for erroreous cases the value won't be checked, so no need to specify

carbon/schemas_test.go
```diff
@@ -0,0 +1,106 @@
...
+				return
+			}
+			if tt.wantName != got.Name {
+				t.Errorf("name %v, want %v", got.Name, tt.wantName)
+			}
+			if len(tt.wantTags) != len(got.Tags) {
+				t.Errorf("tags '%v', want'%v'", got.Tags, tt.wantTags)
+			}
+			for i := range tt.wantTags {
+				t.Logf("i=%v", i)
```
Dieterbe: why the log here? maybe you want to print the i only in the failure clause below?

Dieterbe:
question for @replay , @DanCech do we want the tags to be sorted at this point? I'ld say yes, and let's add a test case for an unsorted-tag carbonline, where we expect the tags to come out sorted

DanCech:
Yes, tags should be sorted

jtlisi:
@Dieterbe The slice of tags is sorted when the metric Id is set.

Dieterbe:
I think as a general rule, we should never merge master into a feature branch, as it makes things harder to diff.
it's trivial to diff a feature branch against whatever version of master it was based on, so there's rarely a need to pull in master stuff into a feature branch. if you really want to do it, better to rebase the feature master branch on top of master and force push

Dieterbe:
please remove the `Merge branch 'master' into update-carbon-tagging` commit (just run `git rebase -i` on the branch and remove the merge commit, then you can force push to origin/update-carbon-tagging), and ping me again when all feedback has been addressed. thanks

jtlisi:
I rebased on master and squashed the changes into a single commit. Hopefully this is sorted

Dieterbe:
I still think (see earlier comment https://github.com/raintank/tsdb-gw/pull/58#discussion_r152398254 ) that e.g. : 

```
+                       wantName: "test.metric.value",
+                       wantTags: nil,
+                       wantErr:  true,

```
is at least misleading, arguably incorrect. Because that's not really the name we want in an error case, and in fact we know that in error cases we return nil as MD so there is no name andwe skip the name checking in that case.

carbon/schemas.go
```diff
@@ -30,7 +30,20 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
...
 
-	name := elements[0]
+	metric := strings.Split(elements[0], ";")
+	name := metric[0]
+
+	var tags []string
+
+	if len(metric) > 1 {
+		for _, v := range metric[1:] {
+			if v != "" && strings.Contains(v, "=") && v[0] != '=' {
```
Dieterbe: this effectively validates that the tag key is non-empty. But I think we can also validate that : 
* tag value is non empty
* that there's only1 equals character

anything i'm missing @DanCech ? any forbidden characters or other restrictions?

carbon/schemas.go
```diff
@@ -30,7 +30,20 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
...
+				tags = append(tags, v)
+			} else {
+				return nil, fmt.Errorf(errFmt, msg, "unable to format tag")
```
Dieterbe: "format tag" implies that we're trying to create a formatted string, which is the opposite of what's going on. should be "unable to parse tag", or "incorrect tag format" or something.

carbon/schemas_test.go
```diff
@@ -0,0 +1,116 @@
...
+			if tt.wantName != got.Name {
+				t.Errorf("name %v, want %v", got.Name, tt.wantName)
+			}
+			if len(tt.wantTags) != len(got.Tags) {
+				t.Errorf("tags '%v', want'%v'", got.Tags, tt.wantTags)
+			}
+			for i := range tt.wantTags {
+				if tt.wantTags[i] != got.Tags[i] {
+					t.Errorf("exptected tag to match, tag=%v, want %v", got.Tags[i], tt.wantTags[i])
+				}
```
Dieterbe: can you make the error formatting consistent. each test case has a name, so why not include the name in every Error scenario?

carbon/schemas_test.go
```diff
@@ -0,0 +1,116 @@
...
+				schemas: &schemas,
+				orgID:   1,
+			},
+			wantName: "test.metric.value",
+			wantTags: nil,
+			wantErr:  true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
Dieterbe: this is neat. i like how it automatically prints the test case name.  we should do this more in other places also.

jtlisi: Agreed, vscode can autogenerate the fixtures for unit tests like this out of the box

carbon/schemas.go
```diff
@@ -30,7 +30,26 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
...
+			tagset := strings.Split(v, "=")
+			if len(tagset) != 2 {
+				return nil, fmt.Errorf(errFmt, msg, "unable to parse tag, more than one '=' character")
+			}
+			if tagset[0] == "" {
+				return nil, fmt.Errorf(errFmt, msg, "unable to parse tag, no key")
+			}
+			if tagset[1] == "" {
+				return nil, fmt.Errorf(errFmt, msg, "unable to parse tag, no value")
+			}
```
Dieterbe: i just learned from Dan that apparently, empty values are legal.
to be more precise he said: 
> technically we just validate `^[^;]+(;[^;=]+=[^;]*)*$`
> TaggedSeries.parse_carbon() will accept anything matching that regex

carbon/schemas.go
```diff
@@ -30,7 +30,26 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
...
 
-	name := elements[0]
+	metric := strings.Split(elements[0], ";")
+	name := metric[0]
+
+	var tags []string
+
+	if len(metric) > 1 {
+		for _, v := range metric[1:] {
+			tagset := strings.Split(v, "=")
```
Dieterbe: this is allocation heavy and slows down our app. I was more a fan of your previous style of looking at position(s) of `=` where we don't allocate temporary slices and strings

carbon/schemas.go
jtlisi: Why that regex? Don't we just have to validate the existence of a key and then ensure the tag contains not equals signs like this: `^[^;=]+=[^;=]*$`?

DanCech: the tag can't contain `=`, the value may

DanCech: https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/tags/utils.py#L40-L59

jtlisi: Gotcha, so `^[^;=]+=[^;]*$` should work

DanCech: I don't think the intention was to use regexp, take a look at the code I linked.  You already know that the string doesn't contain `;` because you split on `;` above so you can just say `strings.SplitN(v, "=", 2)` and validate that you get 2 elements back and the first one isn't empty.

Dieterbe: Yes, the first version of the code you had was actually the best, it was simple and performed the best. this is hot code where we should avoid allocations, regex matching, etc.

jtlisi:
@Dieterbe Can you check this out

carbon/schemas.go
replay: sorry for being late to the discussion. i was under the impression that the value can only be empty in the tag queries (that has a special meaning), but it can not be empty as a part of the metric identifier.
Otherwise, if the value can be empty as a part of the metric id, then i would have to go through the MT tagging stuff again because i don't think an empty value would be accepted currently. 
But anyway, how would that then be queried with `key=` already has the special meaning of the tag `key` not being present?

DanCech: In graphite `key=` and "key not present" are treated the same for querying, the difference is in how the series name is formatted.  If the user sends an explicit `key=` then it is part of the series identifier.

Dieterbe:
as i said before, we need to switch back to the code you had at first where you just iterated the existing slice, definitely not regex matching here.

DanCech:
@jtlisi did you read the comments above? You should.

https://github.com/raintank/tsdb-gw/pull/58#discussion_r154395690

carbon/schemas.go
```diff
@@ -30,7 +30,19 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
...
+	var tags []string
+
+	if len(metric) > 1 {
+		for _, v := range metric[1:] {
+			if v == "" || !strings.Contains(v, "=") || v[0] == '=' {
+				return nil, fmt.Errorf(errFmt, msg, "unable to parse tag")
+			}
+			tags = append(tags, v)
+		}
+	}
```
DanCech: Can't we just say:
```golang
tags := metric[1:]
for _, v := range tags {
  if v == "" || !strings.Contains(v, "=") || v[0] == '=' {
    return nil, fmt.Errorf(errFmt, msg, "unable to parse tag")
  }
}
```

jtlisi: Yea that should work

jtlisi:
Cool, just rebased onto master

carbon/schemas.go
Dieterbe: i don't think it would. if anything this approach may introduce an extra allocation when `len(metric) == 1` but i'm not sure. there's all kinds of compile flags and analysis you can do if you want to get to the bottom of it (probably not worth the time though)

the new approach however would avoid doing a needless branch (jump) by taking out the if (the for loop already automatically runs an if condition similar to the one you had, every time it loops)

carbon/schemas.go
```diff
@@ -30,7 +30,15 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
 		return nil, fmt.Errorf(errFmt3Fields, msg)
 	}
 
-	name := elements[0]
+	metric := strings.Split(elements[0], ";")
+	name := metric[0]
+
+	tags := metric[1:]
+	for _, v := range tags {
+		if v == "" || !strings.Contains(v, "=") || v[0] == '=' {
```
Dieterbe: wouldn't this allow multiple equals signs? should probably replace the `strings.Contains` clause with a `strings.Count` clause

jtlisi: I think multiple equals signs are allowed

DanCech: = is allowed in values

carbon/schemas.go
```diff
@@ -30,7 +30,15 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
...
+			return nil, fmt.Errorf(errFmt, msg, "unable to parse tag")
```
Dieterbe: IMHO when a function has a problem with an input, it should never repeat the entire input in its error message.  if the caller is interested, they can always log the input and the error it got back from the parseMetric function.

a function should thus either log the specific aspect or portion of the input it has a problem with (e.g. the tag field that is invalid), or just return the error. I think the latter suffices here. this frees you up to not have to call Errorf (expensive), anyway we already have the pre-existing errFmt3Fields, after we merge this PR i'll create one with this change to show what I mean and you guys can review

carbon/schemas.go
Dieterbe: my bad, values may have an equals in them.

carbon/schemas.go
jtlisi: That makes sense. Maybe we can create a internal doc that can serve as a basis for a golang style guide

PR #59, opened 2017-11-20
remove worldping specific code

woodsaj:
worldping specific code has been moved to a new service (worldping-gw).


scripts/config/tsdb.ini
```diff
@@ -10,14 +10,6 @@ metrics-kafka-comp = none
 metrics-publish = false
 metrics-partition-scheme = bySeries
 
-events-topic = events
-events-publish = false
-events-compression = none
-
-grahite-url = http://localhost:8080
```
Dieterbe: shouldn't we keep graphite-url?

woodsaj: good catch.  added it back, and also added a default for "metrictank-url"

PR #61, opened 2017-12-13
Misc post carbon tagging

Dieterbe:
feel free to be as harsh on this PR as i've been on the tagging PR ;-)

### DIFF
--- a/carbon/schemas.go
+++ b/carbon/schemas.go
@@ -1,6 +1,7 @@
 package carbon
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


carbon/schemas.go
```diff
@@ -36,18 +37,18 @@ func parseMetric(buf []byte, schemas *conf.Schemas, orgId int) (*schema.MetricDa
 	tags := metric[1:]
 	for _, v := range tags {
 		if v == "" || !strings.Contains(v, "=") || v[0] == '=' {
-			return nil, fmt.Errorf(errFmt, msg, "unable to parse tag")
+			return nil, errBadTag
```
DanCech: Might be good here to tell the user which tag segment was bad

jtlisi: Based on what you said in the previous PR @Dieterbe, would a log statement here with the tag work?

Dieterbe: the caller of parseMetric can (and does) log the entire metric key that was given to parseMetric, when there was an error. I think that's the best trade-off

jtlisi: Ahh, that seems best

PR #63, opened 2018-01-22
Get rid of Sarama and use Confluent

replay:
I commented the building of the docker image on CircleCI because the docker version in Circle 1.0 is too old for the `From .. AS ..` syntax in the updated `Dockerfile`. But we need that because that way the image size is reduced from ~850MB to ~50MB. I think we should update this project to Circle 2.0 soon, then we can reenable the building of the docker image in the circle tests.


metric_publish/publish.go
```diff
@@ -115,34 +86,38 @@ func Publish(metrics []*schema.MetricData) error {
 		if err != nil {
 			return err
 		}
-		payload[i] = &sarama.ProducerMessage{
-			Key:   sarama.ByteEncoder(key),
-			Topic: topic,
-			Value: sarama.ByteEncoder(data),
+
+		msg := &kafka.Message{
+			TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
```
woodsaj: how does kafka.PartitionAny work?  Looking at the code, it doesnt appear to behave the same as Sarama, which is bad.   We need metrics to be partitioned exactly the same.

By default, Sarama uses its own hashPartitioner  https://github.com/Shopify/sarama/blob/master/partitioner.go#L102-L131

The confluent-kafka-go looks to use the value of the "partitioner' topic configuration property which defaults to 'consistent_random'  https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md#topic-configuration-properties

I think we are going to need to set the partition here based on the sarama implementation.

woodsaj: Looks like we can simply get the number of partitions during `Init(broker string)` with a call to
```
meta, err := producer.GetMetadata(topic, false, 30000)
partitionCount = len(meta.Topics[topic].Partitions)
```
Though it is possible to change the number of partitions used by a Topic, we dont do that.  So i dont think we need to periodically call GetMetadata.    But it wouldnt be hard to just create a goroutine with a time.Ticker that makes the call and uses an atomic write to update partitionCount.

With the paritionCount known, we can just implement the FNV1-a hashing that sarama does to explicitly set the partition id of the message

metric_publish/publish.go
```diff
@@ -44,105 +86,91 @@ func init() {
...
 		return
 	}
 	var err error
-	partitioner, err = p.NewKafka(partitionScheme)
+
+	config := kafka.ConfigMap{}
+	config.SetKey("request.required.acks", "all")
+	config.SetKey("message.send.max.retries", "10")
+	config.SetKey("bootstrap.servers", broker)
+	config.SetKey("compression.codec", codec)
```
Dieterbe: where are these compression codec / settings mechanism documented?
can't find on https://godoc.org/github.com/confluentinc/confluent-kafka-go/kafka#ConfigMap.SetKey or https://docs.confluent.io/current/clients/confluent-kafka-go/index.html#ConfigMap.SetKey 
https://cwiki.apache.org/confluence/display/KAFKA/Compression uses integer values instead of strings :?

woodsaj: @Dieterbe confluent-kafka-go is a wrapper for librdkafka.  So you can find the configuration params at https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

scripts/build.sh
```diff
@@ -1,25 +1,33 @@
...
+
+cd $SOURCE_DIR
+source scripts/build_deps.sh
+
+# get git version
+GIT_VERSION=`git describe --always`
 
 # Disable CGO for builds.
-export CGO_ENABLED=0
+export CGO_ENABLED=1
```
Dieterbe: please update the comment to describe what's going on here

metric_publish/publish.go
```diff
@@ -24,17 +25,58 @@ var (
...
 	flushFreq       time.Duration
+	partitioner     Partitioner
 	maxMessages     int
-
-	bufferPool = util.NewBufferPool()
 )
 
+type Partitioner interface {
+	partition(schema.PartitionedMetric) (int32, error)
+}
```
Dieterbe: might make more sense to use MT's `partitioner.Partitioner` interface?

metric_publish/publish_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"testing"
+
+	"github.com/Shopify/sarama"
+	p "github.com/raintank/metrictank/cluster/partitioner"
+	"gopkg.in/raintank/schema.v1"
+)
+
+func TestPartitioning(t *testing.T) {
```
Dieterbe: what are we really testing here? p.Kafka and sarama.HashPartitioner will behave the same because the former relies on the latter, in fact basically the same code that is run here (creating the `key`, `msg` and calling Partition on the sarama HashPartitioner) is used in both implementations

Dieterbe:
can you also rebase this on top of master and cleanup the commit history

metric_publish/publish_test.go
replay: actually this was only a test that i wrote for myself to verify that the reimplementation of the partitioner really does the exact same thing. then i figured it wouldn't hurt to commit it. but you're right that there's not really a point to leaving it committed because it's result most likely won't change

metric_publish/publish.go
```diff
@@ -44,105 +86,91 @@ func init() {
...
 			return err
 		}
-		payload[i] = &sarama.ProducerMessage{
-			Key:   sarama.ByteEncoder(key),
-			Topic: topic,
-			Value: sarama.ByteEncoder(data),
+
+		msg := &kafka.Message{
+			TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: part},
+			Value:          []byte(data),
```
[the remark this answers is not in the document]
replay: Fixed

metric_publish/publish.go
replay: unfortunately that won't work because we also need the key to be returned in this case

metric_publish/publish_test.go
Dieterbe: why not keep it but verify that the new confluent partitioner behaves the same as partitioner.Kafka ?

metric_publish/publish.go
Dieterbe: then why does this function signature not return a key? when i compare the 2 interfaces, they seem compatible, other than that instead of passing numPartitions at creation time of the partitioner, it's when you call the function. other then that, they are identical.

replay: it does now, i added after you wrote that comment

metric_publish/publish_test.go
replay: i'm not sure what you mean, as you already said the new confluent partitioner is relying on `partition.Kafka`

replay: updated

replay:
@Dieterbe that's ready to review again

scripts/Dockerfile
```diff
@@ -1,13 +1,32 @@
-FROM alpine
-MAINTAINER Anthony Woods awoods@raintank.io
+ARG BASE_PATH=/go/src/github.com/raintank/tsdb-gw
+
+# build binaries inside an alpine3.7 container
+FROM golang:1.9.3-alpine3.7 AS build
+ARG BASE_PATH
 
 RUN apk --update add ca-certificates
```
Dieterbe: do we need the ca-certificates in the build image?

scripts/Dockerfile
```diff
@@ -1,13 +1,32 @@
-FROM alpine
-MAINTAINER Anthony Woods awoods@raintank.io
+ARG BASE_PATH=/go/src/github.com/raintank/tsdb-gw
```
Dieterbe: what's the value of making this an ARG parameter? we don't use it in our `docker build` command, and it's only used for paths within the images, so we can just hardcode them, no?

scripts/Dockerfile
```diff
@@ -1,13 +1,32 @@
...
+RUN $BASE_PATH/scripts/build.sh
+
+
+# now build the production image based on the same alpine version
+# as was used to build the binaries in
+FROM alpine:3.7
+ARG BASE_PATH
+
+RUN apk --update add ca-certificates
+RUN apk add openssl cyrus-sasl
```
Dieterbe: why do we need these two ?

scripts/Dockerfile
```diff
@@ -1,13 +1,32 @@
...
+
 RUN mkdir -p /etc/raintank
-COPY config/tsdb.ini /etc/raintank/tsdb.ini
+COPY scripts/config/tsdb.ini /etc/raintank/tsdb.ini
+COPY scripts/config/tsdb.ini /etc/raintank/tsdb.ini
```
Dieterbe: 2 times same command?

scripts/Dockerfile
replay: we could just hard code them. this would only make it more convenient if we'd want to change it, but there's no real difference if we hardcode it or not.

scripts/Dockerfile
replay: because `librdkafka` requires them

scripts/build.sh
```diff
@@ -1,25 +1,38 @@
...
 
-# Disable CGO for builds.
-export CGO_ENABLED=0
+cd $SOURCE_DIR
+
+if ! [ -d $PKG_CONFIG_PATH ] || [ -z $PKG_CONFIG_PATH ]
+then
+	source scripts/build_deps.sh
+else
+	echo "not building librdkafka"
```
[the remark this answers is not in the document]
replay: when developing and calling `scripts/build.sh` to test it this saves a lot of time because it prevents the `librdkafka` needs to be rebuilt each time.

scripts/Dockerfile
replay: that's a mistake, fixing it

scripts/Dockerfile
replay: I'm actually not sure why they are there, i just left them because they were there already. I'd guess it might be necessary to have them in an unusual case where tsdb-gw needs to connect to a backend via https

Dieterbe: > in the build image

replay: ah right, i dont think we need them in the build image

metric_publish/publish.go
```diff
@@ -44,105 +86,92 @@ func init() {
...
 		return nil
 	}
 	var err error
-
-	payload := make([]*sarama.ProducerMessage, len(metrics))
 	pre := time.Now()
+	deliveryChan := make(chan kafka.Event)
 
-	for i, metric := range metrics {
-		data := bufferPool.Get()
```
Dieterbe: why did you take out the bufferPool ?

metric_publish/publish.go
```diff
@@ -44,105 +86,92 @@ func init() {
...
 
-	// We are looking for strong consistency semantics.
-	// Because we don't change the flush settings, sarama will try to produce messages
-	// as fast as possible to keep latency low.
-	config := sarama.NewConfig()
-	config.Producer.RequiredAcks = sarama.WaitForAll // Wait for all in-sync replicas to ack the message
-	config.Producer.Retry.Max = 10                   // Retry up to 10 times to produce the message
-	config.Producer.Compression = getCompression(codec)
-	config.Producer.Return.Successes = true
-	config.Producer.Flush.Frequency = flushFreq
```
Dieterbe: we're not setting this anymore?

metric_publish/publish.go
```diff
@@ -44,105 +86,92 @@ func init() {
...
 	}
 
-	// We are looking for strong consistency semantics.
-	// Because we don't change the flush settings, sarama will try to produce messages
-	// as fast as possible to keep latency low.
-	config := sarama.NewConfig()
-	config.Producer.RequiredAcks = sarama.WaitForAll // Wait for all in-sync replicas to ack the message
-	config.Producer.Retry.Max = 10                   // Retry up to 10 times to produce the message
-	config.Producer.Compression = getCompression(codec)
-	config.Producer.Return.Successes = true
```
Dieterbe: we're no longer setting this option. why? does it happen automatically or something?

metric_publish/publish.go
```diff
@@ -44,105 +86,92 @@ func init() {
...
-	config.Producer.Flush.Frequency = flushFreq
-	config.Producer.Flush.MaxMessages = maxMessages
```
Dieterbe: no longer setting this?

metric_publish/publish.go
replay: in the old version with sarama a slice of payloads was first built and then passed to sarama to process in batch, so the buffer pool saved quite a number of allocations. in the new version with confluent batch processing exists, but it is marked as unstable, so i didn't use it. since batch processing isn't used anymore we only need to allocate one `data` object per call to `.Publish()` because this same object can be re-used when ranging over the given `metrics` instead of allocating a whole slice of them, so the number of allocations saved by using the buffer pool would have decreased a lot which made it seem like an unnecessary complication.

metric_publish/publish.go
woodsaj: This is the frequency to publish batches, which we arent doing anymore

metric_publish/publish.go
woodsaj: This is the max number of messages to publish per batch. We are not sending batches anymore.
But we might want to set `queue.buffering.max.messages` and  `max.in.flight.requests.per.connection` .  Though not exactly the same, they will have similar impact on publish throughput.

metric_publish/publish.go
woodsaj: This is a sarama internal channel.  confluent-kafka-go doesnt have an equivalent

metric_publish/publish.go
woodsaj: This is the interval to flush each batch. confluent-kafka-go doesnt batch messages

metric_publish/publish_test.go
```diff
@@ -0,0 +1,39 @@
...
+	part_old, _ := p.NewKafka("bySeries")
+
+	var data []byte
+	for partitions := int32(1); partitions <= 64; partitions++ {
+		part_new := NewPartitioner(partitions)
+
+		for _, m := range testData {
+			data, _ = m.MarshalMsg(data)
+			res_old, _ := part_old.Partition(&m, partitions)
+			res_new, _, _ := part_new.partition(&m)
```
Dieterbe: can you also verify that both return the same key? (should be very clearly correct, since part_new uses part_old to get the key, but still)

metric_publish/publish.go
```diff
@@ -44,105 +86,92 @@ func init() {
...
-			}
-		} else {
+	}
+
+	msgCount := 0
+	for e := range deliveryChan {
+		msgCount++
+
+		m, ok := e.(*kafka.Message)
+		if !ok || e == nil {
```
Dieterbe: interestingly, the official example doesn't seem to handle this case.
https://github.com/confluentinc/confluent-kafka-go/blob/master/examples/producer_example/producer_example.go#L54

metric_publish/publish.go
```diff
@@ -44,105 +86,92 @@ func init() {
...
+			sendErrOther.Inc()
+		} else if m.TopicPartition.Error != nil {
 			sendErrOther.Inc()
-			log.Error(4, "SendMessages error: %s", err.Error())
+			log.Error(4, "SendMessages error%v\n", m.TopicPartition.Error)
+		}
```
Dieterbe: in both of these failure cases, shouldn't we set some error to return?
after this has run and we close the channel, we will return a `nil` error, acknowledging the writes.
this function may only return a nil error if we safely sent everything into kafka

Dieterbe: also i don't think we should log an error for every single message that failed.
if there's a temporary failure affecting all messages, then this would log a full line for every single datapoint, which is way too much.

instead, maybe print the first error and a count of the number of errors

metric_publish/publish.go
Dieterbe: the proposed code here allocates a new `[]byte` for every metric, though.
I doubt you can do what you mention, because the documentation for Produce says:
```
// This is an asynchronous call that enqueues the message on the internal
// transmit queue, thus returning immediately.
```
so if you were to reuse a slice immediately after calling Produce, you'd overwrite the message it's still working on enqueueing.

it looks to me like we must stick to the previous approach of using distinct slices for each message within one produce call. once we got the confirmation they were all shipped off successfully, only then can we reuse the slices (and the bufferPool should help with that)

metric_publish/publish_test.go
replay: afaict that would require me to modify metrictank's `cluster/partitioner:Kafka`, so I'm not sure that's worth it

metric_publish/publish.go
replay: it should be fine because the payload gets copied when en-queueing https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_msg.c?#L117

but if you want i can change it back anyway

replay: it should be fine because with the flag `RD_KAFKA_MSG_F_COPY` the payload gets copied here https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_msg.c#L117 and the flag gets set in the confluent lib https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/producer.go#L160

but if you want i'm fine to change it back anyway

metric_publish/publish.go
replay: yeah, but it's still better to do so or not?

metric_publish/publish.go
replay: good point, fixing

replay: how's that https://github.com/raintank/tsdb-gw/pull/63/commits/bb172e98e2a03f3991b2b9b9979e818c74df1ec0

metric_publish/publish_test.go
Dieterbe: it has a `GetPartitionKey` method. you're calling it in the new code

metric_publish/publish.go
replay: added that https://github.com/raintank/tsdb-gw/pull/63/commits/966812dfd79d19e2baef3808590887db1c80f23a

metric_publish/publish_test.go
replay: ah, ok. i thought you'd want me to get the actual key that's used inside `.Partition()`

replay: https://github.com/raintank/tsdb-gw/pull/63/commits/ea6d66b6e904ca98b1133467ad5b205742fcd8e1

metric_publish/publish.go
Dieterbe: i got a bit lost in the C code but if you're sure the function blocks until the memory is safely copied, then that's fine, but then still , as i said above, it looks like you instantiate a new buffer in each iteration.

Dieterbe:
maybe i'm stating the obvious here, but as a reminder: the tsdbgw->kafka ingestion path is one of the most critical parts of our infrastructure. we have to be 100% confident that when we ack an ingest request to a customer, that the data has been persisted safe and sound into kafka.
so any accidental memory overwriting, improper ack / error handling, decode failures, etc are to be avoided at all costs. the code looks generally good, but we have to avoid any surprised.

we should think a bit more about how to test this. i'm not sure how. but maybe a highly concurrent workload to the ingest point (can be done with fakemetrics), and then asserting all data comes out properly out of kafka? not the most practical and representative test though. can we mock kafka somehow?

metric_publish/publish.go
Dieterbe: yes. in fact the type assertion didn't work, and e is not nil, i would log an error like `fmt.Errorf("unexpected delivery report of type %T: %v", e, e)` , just make sure to log that error but not report it back to the customer as it may contain too much information

metric_publish/publish.go
replay: somehow i've just assumed that go would be smart enough to reuse the same slice of bytes if i declare `var abc []byte` inside a loop, but i think you might be right i'll fix that.

replay: you're right that we shouldn't take any risks with the ingest path, so let's go back to the old method with the buffer pool then

replay: https://github.com/raintank/tsdb-gw/pull/63/commits/66c1d8e521a9da8c05f34ad63f6b4c6ee75bf4df

replay:
rebased & merging

PR #65, opened 2018-01-25
Upgrade to CircleCI 2.0

replay:
This upgrades the build to CircleCI 2.0


circle.yml
```diff
@@ -1,37 +0,0 @@
...
-general:
-  artifacts:
-    - build
-deployment:
-  production:
-    branch: master
-    commands:
-      - scripts/deploy.sh
-      - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
-      - scripts/deploy_docker.sh
```
Dieterbe: are we not calling any of the deploy stuff anymore?

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+          test -e download/$GODIST || curl -o download/$GODIST https://storage.googleapis.com/golang/$GODIST
+          sudo rm -rf /usr/local/go
+          sudo tar -C /usr/local -xzf download/$GODIST
+          sudo ln -s /usr/local/go/bin/* /bin/
+    - run:
+        name: run tests
+        command: |
+          scripts/vendor_health.sh
+          scripts/build.sh
+          scripts/build_docker.sh
```
Dieterbe: why are we building in a step called "run tests" ?

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+          go test -v -race $(go list ./... | grep -v /vendor/)
+          go vet $(go list ./... | grep -v /vendor/)
```
Dieterbe: why not switch to the latest release (1.9.2 or 1.9.3 if it's available on circle), then the vendor exclusion will happen automatically.

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+      GODIST: go1.8.1.linux-amd64.tar.gz
+    steps:
+    - checkout
+    - run:
+        name: prepare build env
+        command: |
+          sudo apt-get update
+          sudo apt-get install -y curl git bundler rpm
+          mkdir -p download
+          test -e download/$GODIST || curl -o download/$GODIST https://storage.googleapis.com/golang/$GODIST
```
Dieterbe: it would be nice if we didn't have to manually install go.
https://circleci.com/docs/2.0/executor-types/#using-machine has a link to the script they use to generate their images. since it's still stuck with 1.7.3 :/

Dieterbe: anyway just using a docker executor with recent golang image would also make this much easier.
presumably the issue you saw with setup_remote_docker should be fixed now

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
+version: 2
+jobs:
+  build:
+    machine:
```
woodsaj: use docker images, not a machine.  See metrictank, worldping-gw or raintank-probe for examples of circleci2.0 configs.

replay: fixing. the last time i tried there was some problem in circle ci which made the `setup_remote_docker` step break with an internal `503`, but that seems to be working now.

.circleci/config.yml
replay: doing that now

.circleci/config.yml
replay: done

.circleci/config.yml
```diff
@@ -0,0 +1,34 @@
+version: 2
+jobs:
+  build:
+    docker:
+    - image: circleci/golang:1.9.3
+    working_directory: /home/circleci/.go_workspace/src/github.com/raintank/tsdb-gw
+    parallelism: 1
+    shell: /bin/sh
+    environment:
```
woodsaj: You dont need these ENV vars

circle.yml
replay: fixed

.circleci/config.yml
replay: fixed

replay:
@woodsaj @Dieterbe used raintank-probe as a template now and added the `vendor_health.sh` to it.

.circleci/config.yml
replay: 👍

.circleci/config.yml
```diff
@@ -0,0 +1,67 @@
...
+
+  test:
+    working_directory: /go/src/github.com/raintank/tsdb-gw
+    docker:
+      - image: circleci/golang:1.9.3
+    steps:
+      - checkout
+      - run:
+          name: unit tests
+          command: go test -v -race $(go list ./...)
```
Dieterbe: can simplify to `go test -v -race ./..` same for the `go vet` command

circle.yml
```diff
@@ -1,37 +0,0 @@
...
-    - scripts/package.sh
-dependencies:
-  override:
-    - cd $IMPORTPATH && scripts/depends.sh
-    - cd $IMPORTPATH && scripts/vendor_health.sh
-    - cd $IMPORTPATH && scripts/build.sh
-    - cd $IMPORTPATH && scripts/build_docker.sh
-general:
-  artifacts:
-    - build
```
Dieterbe: so the new one no longer saves the artifacts. I think that's mostly fine except sometimes it can be handy to download a generated binary from a specific build.

.circleci/config.yml
```diff
@@ -0,0 +1,67 @@
...
+  deploy:
+    docker:
+      - image: circleci/ruby:2.3
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - setup_remote_docker
+      - run:
+          name: install dependencies for buidling packages
```
Dieterbe: typo 'buidling'
but actually, i would just use the simplified syntax of writing ` -run : bundle install` (similar for all other run commands), that way this file becomes much more compact, and another benefit is in circleci you then immediately see all the commands run, instead of having to expand on the names. our commands tend to be self-descriptive and just as clear as the names we give them.

circle.yml
replay: readded that

scripts/vendor_health.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+if ! echo $PATH | grep $GOPATH/bin >/dev/null; then
```
Dieterbe: can simplify to `if ! grep -q $GOPATH/bin <<< $PATH; then`

.circleci/config.yml
```diff
@@ -0,0 +1,55 @@
...
+    docker:
+      - image: circleci/ruby:2.3
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - setup_remote_docker
+      - run: bundle install
+      - run: scripts/build_docker.sh
+      - run: scripts/package.sh
```
Dieterbe: if you move the store_artifacts stuff here then it will also store the packages as artifacts. only rarely useful, but would be good for consistency with MT

replay: pushed again @Dieterbe

PR #67, opened 2018-02-08
Prometheus functionality

jtlisi:
This is a cleaned up version of tsdb-gw with the prometheus code behind the macaron handler so basic auth can be used. I tested it locally but not extensively.


jtlisi:
Let me rebase this before review

api/api.go
```diff
@@ -95,6 +95,7 @@ func (a *Api) InitRoutes(m *macaron.Macaron) {
 	m.Get("/metrics/index.json", a.Auth(), MetrictankProxy("/metrics/index.json"))
 	m.Get("/graphite/metrics/index.json", a.Auth(), MetrictankProxy("/metrics/index.json"))
 	m.Any("/graphite/*", a.Auth(), GraphiteProxy)
+	m.Any("/prometheus/receive", a.Auth(), PrometheusWrite)
```
DanCech: not sure whether we should make this `/prometheus/write` since we're planning to add `/prometheus/read` later

woodsaj: I also prefer the term "write" here.

api/prometheus.go
```diff
@@ -0,0 +1,97 @@
...
+				if l.Name == model.MetricNameLabel {
+					name = l.Value
+				} else {
+					tagSet = append(tagSet, l.Name+"="+l.Value)
+				}
+			}
+			if name != "" {
+				for _, sample := range ts.Samples {
+					md := metricPool.Get()
+					*md = schema.MetricData{
```
woodsaj: You need to set the `Interval` field.

api/prometheus.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/raintank/tsdb-gw/metricpool"
+	"github.com/raintank/worldping-api/pkg/log"
+	schema "gopkg.in/raintank/schema.v1"
+)
+
+var (
+	metricPool = metricpool.NewMetricDataPool()
+)
+
+func getSchemas(file string) (*conf.Schemas, error) {
```
woodsaj: I dont see this getting called by anything.

Dieterbe:
```
api/prometheus.go:7:2: cannot find package "github.com/gogo/protobuf/proto" in any of:
	/go/src/github.com/raintank/tsdb-gw/vendor/github.com/gogo/protobuf/proto (vendor tree)
	/usr/local/go/src/github.com/gogo/protobuf/proto (from $GOROOT)
	/go/src/github.com/gogo/protobuf/proto (from $GOPATH)
api/prometheus.go:9:2: cannot find package "github.com/grafana/metrictank/conf" in any of:
	/go/src/github.com/raintank/tsdb-gw/vendor/github.com/grafana/metrictank/conf (vendor tree)
	/usr/local/go/src/github.com/grafana/metrictank/conf (from $GOROOT)
	/go/src/github.com/grafana/metrictank/conf (from $GOPATH)
api/prometheus.go:10:2: cannot find package "github.com/prometheus/common/model" in any of:
	/go/src/github.com/raintank/tsdb-gw/vendor/github.com/prometheus/common/model (vendor tree)
	/usr/local/go/src/github.com/prometheus/common/model (from $GOROOT)
	/go/src/github.com/prometheus/common/model (from $GOPATH)
api/prometheus.go:11:2: cannot find package "github.com/prometheus/prometheus/prompb" in any of:
	/go/src/github.com/raintank/tsdb-gw/vendor/github.com/prometheus/prometheus/prompb (vendor tree)
	/usr/local/go/src/github.com/prometheus/prometheus/prompb (from $GOROOT)
	/go/src/github.com/prometheus/prometheus/prompb (from $GOPATH)
Exited with code 1
```

scripts/deploy_docker.sh
```diff
@@ -8,4 +8,4 @@ cd ${DIR}
 VERSION=`git describe --always`
 
 docker push raintank/tsdb-gw:$VERSION
-docker push raintank/tsdb-gw:latest
+# docker push raintank/tsdb-gw:latest
```
Dieterbe: what does this have to do with prometheus?

api/prometheus.go
```diff
@@ -0,0 +1,112 @@
...
+	var err error
+	flag.StringVar(&schemasConf, "prom-schemas-file", "/etc/storage-schemas.conf", "path to carbon storage-schemas.conf file for prom metrics")
+
+	schemas, err = getSchemas(schemasConf)
+	if err != nil {
+		log.Fatal(4, "failed to load schemas config. %s", err)
+	}
+}
+
+func getSchemas(file string) (*conf.Schemas, error) {
```
Dieterbe: this function wrappes conf.ReadSchemas unnessarily. can just call conf.ReadSchemas directly instead of calling this function

api/prometheus.go
```diff
@@ -0,0 +1,112 @@
...
+	schemas, err := conf.ReadSchemas(file)
+	if err != nil {
+		return nil, err
+	}
+	return &schemas, nil
+}
+
+func PrometheusWrite(ctx *Context) {
+	defer ctx.Req.Request.Body.Close()
+	if ctx.Req.Request.Body != nil {
```
Dieterbe: put the defer after this check, otherwise can get nil panic.

api/prometheus.go
```diff
@@ -0,0 +1,112 @@
...
+						Time:     (sample.Timestamp / 1000),
+						Mtype:    "gauge",
+						Tags:     tagSet,
+						OrgId:    ctx.OrgId,
+					}
+					md.SetId()
+					buf = append(buf, md)
+				}
+			} else {
+				log.Warn("prometheus metric received with empty name")
```
Dieterbe: what does it mean when this happens? did the sender construct invalid data? if so, maybe a http 400 is in order?

jtlisi: The name field of a prometheus metric can not be blank. I guess a 400 would make sense. However, does 1 poorly formatted metric require throwing out the entire payload?

Dieterbe: it's IMHO the best we can do. better at least than dropping the data and not telling the client.

jtlisi: agreed, maybe a counter is in order here

api/prometheus.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func PrometheusInit() {
+	if !*enabled {
+		return
+	}
+	log.Info("prometheus input enabled")
+	s, err := conf.ReadSchemas(*schemaFile)
+	if err != nil {
+		log.Fatal(4, "failed to load prometheus schemas config. %s", err)
```
Dieterbe: either turn %s into %q or first do a check to confirm `*schemaFile != ""`
otherwise this would print only some whitespace and it'll be confusing

api/prometheus.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/raintank/tsdb-gw/metricpool"
+	"github.com/raintank/worldping-api/pkg/log"
+	schema "gopkg.in/raintank/schema.v1"
+)
+
+var (
+	metricPool = metricpool.NewMetricDataPool()
+	schemas    *conf.Schemas
+	schemaFile = flag.String("prom-schemas-file", "/etc/storage-schemas.conf", "path to carbon storage-schemas.conf file for prom metrics")
+	enabled    = flag.Bool("prometheus-enabled", false, "enable prometheus input")
```
Dieterbe: consistency. either prom or prometheus for both

api/prometheus.go
```diff
@@ -90,7 +93,9 @@ func PrometheusWrite(ctx *Context) {
 					buf = append(buf, md)
 				}
 			} else {
-				log.Warn("prometheus metric received with empty name")
+				log.Error(3, "prometheus metric received with empty name: %v", ts.String())
+				ctx.JSON(400, "metric written without name label")
```
Dieterbe: not "written", that's confusing

api/prometheus.go
```diff
@@ -16,17 +16,20 @@ import (
...
 
 var (
-	metricPool = metricpool.NewMetricDataPool()
-	schemas    *conf.Schemas
-	schemaFile = flag.String("prom-schemas-file", "/etc/storage-schemas.conf", "path to carbon storage-schemas.conf file for prom metrics")
-	enabled    = flag.Bool("prometheus-enabled", false, "enable prometheus input")
+	metricPool             = metricpool.NewMetricDataPool()
+	schemas                *conf.Schemas
+	schemaFile             = flag.String("prometheus-schemas-file", "/etc/storage-schemas.conf", "path to carbon storage-schemas.conf file for prom metrics")
+	prometheusWriteEnabled = flag.Bool("prometheus-enabled", false, "enable prometheus input")
```
Dieterbe: is there a specific reason to make this configurable / default to false?
i personally don't really care but i don't see why we wouldn't just accept it by default. that's also what we discussed in the HM meeting on monday.

jtlisi: I have no reason

Dieterbe: ok so switch it to true by default then.

jtlisi: One flag that I want to alter is the prometheus-schema-file. Since it technically is using the same file as graphite I would prefer them to use the same flag if that's ok

Dieterbe: it's conceivable a client may have different schemas for carbon as for prometheus, so i would keep them separate.

Dieterbe:
after publishing the data, you need to put the md's back into the metricPool.
(see carbon input)

api/prometheus.go
```diff
@@ -0,0 +1,112 @@
...
+			log.Error(3, "Unmarshal Error, %v", err)
+			return
+		}
+
+		buf := make([]*schema.MetricData, 0)
+		for _, ts := range req.Timeseries {
+			var name string
+			var tagSet []string
+
+			_, s := schemas.Match(name, 0)
```
Dieterbe: you're doing schemas.Match on name when it is still "" instead of after setting it to the proper value?

Dieterbe:
I fixed the remaining bugs that I saw (please have a look at them). looks good to me now.

PR #68, opened 2018-02-22
tune send buffer

replay:
This improves the performance by a lot

Explanation of the screenshot:

- Up until 18:10 it running the old tsdb-gw with sarama
- At 18:10 I switched it to the `0.8.0` and the metric publishing time exploded
- 18:26 I deployed this branch with default settings and performance improved
- 18:31 I've increased the `metrics-buffer-max-ms` setting to `1000` and performance looked about the same like with Sarama

screenshot from 2018-02-22 18-34-29

### DIFF
--- a/metric_publish/publish.go
+++ b/metric_publish/publish.go
@@ -31,7 +31,6 @@ var (
 	codec           string
 	enabled         bool
 	partitionScheme string
-	flushFreq       time.Duration
 	partitioner     Partitioner
 	maxInFlight     int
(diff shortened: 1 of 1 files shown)


Dieterbe:
does this change the ack semantics? 
IOW: is it still true that we only ack when we have successfully commited into kafka?

woodsaj:
> does this change the ack semantics?

no.  We still only ack once all metrics have been accepted by ALL InSync Replicas.

PR #69, opened 2018-02-22
add prometheus proxy

jtlisi:
This adds the proxy to push requests for Prometheus queries to Metrictank

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -99,6 +99,7 @@ func (a *Api) InitRoutes(m *macaron.Macaron) {
 	if *prometheusWriteEnabled {
 		m.Any("/prometheus/write", a.Auth(), PrometheusWrite)
 	}
+	m.Any("/prometheus/*", a.Auth(), PrometheusProxy)
 }
 
 func index(ctx *macaron.Context) {
--- a/api/prometheus.go
+++ b/api/prometheus.go
@@ -11,6 +11,7 @@ import (
 	"github.com/prometheus/prometheus/prompb"
 	"github.com/raintank/tsdb-gw/metric_publish"
 	"github.com/raintank/tsdb-gw/metricpool"
+	"github.com/raintank/tsdb-gw/metrictank"
 	"github.com/raintank/worldping-api/pkg/log"
 	schema "gopkg.in/raintank/schema.v1"
 )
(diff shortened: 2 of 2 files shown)


api/api.go
```diff
@@ -99,6 +99,7 @@ func (a *Api) InitRoutes(m *macaron.Macaron) {
 	if *prometheusWriteEnabled {
 		m.Any("/prometheus/write", a.Auth(), PrometheusWrite)
 	}
+	m.Any("/prometheus/*", a.Auth(), PrometheusProxy)
```
Dieterbe: i recommend you instead do:
```
m.Get("/prometheus/write", a.Auth(), MetrictankProxy("/prometheus/write")
```

because:
1) more explicit = more clear
2) no need for extra PrometheusProxy function
3) the naming convention being used here is that we name the proxy after the service being routed to
4) no need to construct proxy http path dynamically

downside: will need a few more commits to keep up to date, but it's worth it

woodsaj: that doesnt make sense @Dieterbe 
a) requests sent to "/prometheus/write" are POST
b) requests sent to "/prometheus/write" and handled locally by tsdb-gw. they are not proxied

So what needs to be proxied to Metrictank is "ANY /prometheus/*"  but the MetrictankProxy can't do that.  It needs specific urls.

jtlisi: Does the current MetrictankProxy pass query params?

jtlisi: Actually I just realized we can't add explicit routes because there are wildcards in the prometheus api url.

Dieterbe: can you link to the docs re those wildcards or give an example?

woodsaj: Lets get this merged with the current prometheusProxy approach, which matches the approach used by the graphiteProxy.    Though we currently want the proxy queries through to metrictank we may not always want to do that.

but lets open an issue to clean this up and reduce duplicate code.  
I think both graphiteProxy, MetrictankProxy and PrometheusProxy should all use a standard BackendProxy{} .

eg. 
```
graphiteProxy := NewBackendProxy(GraphiteUrl)
prometheusProxy := newBackendProxy(PrometheusUrl)
```

woodsaj: @Dieterbe 

https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L179

Dieterbe: how is that a problem? for that URL we can just add a proxy route for "/label/" as far as i can see.
but anyway yeah let's just merge as is for now, to save precious pre-grafanacon time.

PR #74, opened 2018-03-09
Partition id fix

replay:
- Reverts the last commit on master which reverted the confluent stuff
- Fixes the partition id generation bug by using separate hashers per thread, instead of only using one global one


metric_publish/publish.go
```diff
@@ -111,38 +150,69 @@ func Publish(metrics []*schema.MetricData) error {
...
+
 		messagesSize.Value(len(data))
+
+		err = producer.Produce(payload[i], deliveryChan)
+		if err != nil {
+			return err
+		}
 	}
+
+	partitionerPool.Put(partitioner)
```
jtlisi: Should this be returned to the pool in the case of an error? Or let the GC take care of it

replay: I think in the case of any error it's probably better to not put it back into the pool. If the pool does not contain enough instances to satisfy all the threads requesting one it will just instantiate a new one.

jtlisi: :+1:

metric_publish/publish.go
```diff
@@ -24,85 +28,120 @@ var (
...
-	bufferPool = util.NewBufferPool()
+	topic            string
+	codec            string
+	enabled          bool
+	partitionScheme  string
+	maxInFlight      int
+	bufferMaxMs      int
+	batchNumMessages int
+	partitionCount   int32
+	kafkaPartitioner *p.Kafka
```
Dieterbe: it took me a little while to figure out why we need this. a comment would be helpful

metric_publish/publish_test.go
```diff
@@ -0,0 +1,45 @@
...
+		{Name: "name7"},
+		{Name: "name8"},
+		{Name: "name9"},
+		{Name: "name10"},
+	}
+
+	part_old, _ := p.NewKafka("bySeries")
+
+	var data []byte
+	kafkaPartitioner, _ = p.NewKafka(partitionScheme)
```
Dieterbe: this is very brittle because part_old uses hardcoded `bySeries` and this one relies on the variable using its default value which is defined in the flags section.

looks like you can use the same kafkaPartitioner for part_old and part_new which is much cleaner.

metric_publish/publish.go
replay: added

metric_publish/publish_test.go
replay: fixed

PR #75, opened 2018-03-16
add opentsdb ingest endpoint with output to kafka

jtlisi:


opentsdb/models.go
```diff
@@ -0,0 +1,18 @@
...
+	Metric    string            `json:"metric"`
+	Timestamp int64             `json:"timestamp"`
+	Value     float64           `json:"value"`
+	Tags      map[string]string `json:"tags"`
+}
+
+type PutRequest []Metric
+
+func (m Metric) FormatTags() []string {
+	tagArray := make([]string, 0)
```
Dieterbe: since you're using the metricPool to recycle `schema.MetricData` you might as well recycle their tag arrays instead of allocating new ones here.
you can do this by giving this function an arg `buf []string`, then append into that buf and return it.
then from the calling code call `Tags:     ts.FormatTags(md.Tags),` and when you put into the pool, use this:
```
m.Tags = m.Tags[:0]
metricPool.Put(m)
```

this way you make sure that any MetricData that comes out of the pool has a 0-length slice for tags (but typically backed by an array that was used previously), so you don't see tags from a previous use.

api/opentsdb.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		var req opentsdb.PutRequest
+		err = json.Unmarshal(body, &req)
+		if err != nil {
+			ctx.JSON(400, err.Error())
+			log.Error(3, "Read Error, %v", err)
+			return
+		}
+		buf := make([]*schema.MetricData, 0)
```
Dieterbe: `var buf []*schema.MetricData`
see https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful or https://go-proverbs.github.io/  click on "Make the zero value useful."

really minor point but anyway

jtlisi:
@Dieterbe this should be good now

PR #76, opened 2018-03-20
cortex support

### LINKED ISSUE #70: add support for cortex
moving forward we want to be able to support using both metrictank and cortex as a backend TSDB.

This allows us to continue to use tsdb-gw for handling auth, then allow metrics and queries through to either backend.

jtlisi:
Super basic support for cortex. All requests to the `/cortex/` path are forwarded to the cortex url. Writes are handled by a dedicated proxy to reduce the number of Allocs.

To Do
- [ ] BufferPool for write reverse proxy

Fixes #70


main.go
```diff
@@ -39,6 +40,8 @@ var (
 	graphiteUrl   = flag.String("graphite-url", "http://localhost:8080", "graphite-api address")
 	metrictankUrl = flag.String("metrictank-url", "http://localhost:6060", "metrictank address")
 
+	cortexUrl = flag.String("cortex-url", "http://localhost:9000", "cortex address")
```
tomwilkie: Cortex has separate read and writes paths, so you'll need flags for both.

api/cortex.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"strconv"
+
+	"github.com/raintank/tsdb-gw/cortex"
+)
+
+func CortexProxy(c *Context) {
+	proxyPath := c.Params("*")
+	proxy := cortex.Proxy(c.OrgId, proxyPath)
```
tomwilkie: Creating a new reverse proxy for each requests seems... suboptimal.  Can we make this the similar to the write path?

api/api.go
```diff
@@ -96,10 +96,13 @@ func (a *Api) InitRoutes(m *macaron.Macaron) {
...
 	m.Any("/graphite/*", a.Auth(), GraphiteProxy)
-	if *prometheusWriteEnabled {
-		m.Any("/prometheus/write", a.Auth(), PrometheusWrite)
+	if *prometheusMTWriteEnabled {
+		m.Any("/prometheus/write", a.Auth(), PrometheusMTWrite)
 	}
 	m.Any("/prometheus/*", a.Auth(), PrometheusProxy)
+
+	m.Any("/cortex/api/prom/push", a.Auth(), CortexWrite)
+	m.Any("/cortex/*", a.Auth(), CortexProxy)
```
tomwilkie: Its advisable not to proxy `*` to Cortex, as this could expose the pprof and metrics endpoints to the public.

tomwilkie: We just proxy everything under `/cortex/api/*` to `/api/*` on the read path.

Also, do we want to use the name `cortex` in the API?  Can we just use the name `prom` or `prometheus`.  At Kausal we use `/prom/api/...`.

tomwilkie:
@jtlisi thanks for this!  I've left a few comments, let me know if you have any questions.

api/middleware.go
```diff
@@ -92,6 +92,17 @@ func (a *Api) Auth() macaron.Handler {
 	}
 }
 
+func (a *Api) CortexHeader() macaron.Handler {
+	return func(ctx *Context) {
+		if ctx.OrgId > 0 {
+			ctx.Req.Request.Header.Add("X-Scope-OrgID", strconv.FormatInt(int64(ctx.OrgId), 10))
+		} else {
+			log.Error(3, "org %v does not exist", ctx.OrgId)
```
Dieterbe: I think by convention we should only log errors for things that are wrong on our side.
if users send bad requests, it's their own fault and not an error that we should log. 

i want to get to the point where we can retrieve all the error logs in our logging platform and see actionable things that we can fix. cc @tomwilkie

woodsaj: @Dieterbe this would be an internal error.  OrgId should be set by the authentication handler.  If it is not, we definitely want to know.

api/middleware.go
```diff
@@ -92,6 +92,17 @@ func (a *Api) Auth() macaron.Handler {
 	}
 }
 
+func (a *Api) CortexHeader() macaron.Handler {
+	return func(ctx *Context) {
+		if ctx.OrgId > 0 {
+			ctx.Req.Request.Header.Add("X-Scope-OrgID", strconv.FormatInt(int64(ctx.OrgId), 10))
```
woodsaj: see https://github.com/raintank/tsdb-gw/issues/70#issuecomment-374356122  this should not be using OrgId.  It needs to use the hosted-metrics instance Id from grafana.com

api/middleware.go
Dieterbe: in that case, we shouldn't tell the user http 400

jtlisi:
This should be up to specs.

cortex/cortex.go
```diff
@@ -37,7 +42,7 @@ func Init() error {
 			req.URL.Scheme = CortexWriteURL.Scheme
 			req.URL.Host = CortexWriteURL.Host
 		},
-		// BufferPool: (add BufferPool for perf reasons if needed)
+		BufferPool: bpool.NewBytePool(*cortexWriteBPoolSize, *cortexWriteBPoolWidth),
```
Dieterbe: why do you create this pool but do nothing with it?

cortex/cortex.go
```diff
@@ -5,12 +5,17 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+
+	"github.com/oxtoacart/bpool"
```
Dieterbe: why this over sync.Pool

Dieterbe:
why `gAuth` package ? packages should not have capital letters. https://blog.golang.org/package-names

api/api.go
```diff
@@ -100,7 +100,13 @@ func (a *Api) InitRoutes(m *macaron.Macaron) {
 		m.Any("/prometheus/write", a.Auth(), PrometheusMTWrite)
 	}
 	m.Any("/prometheus/*", a.Auth(), PrometheusProxy)
+<<<<<<< HEAD
```
Dieterbe: code is more often read than written, so please try to keep the git history a bit clean without merge conflicts. (can squash this commit and the one wherein you resolve the conflict together with `git rebase -i master`

cortex/cortex.go
jtlisi: The httputil.ReverseProxy will take advantage of the pool when proxying requests.
https://golang.org/pkg/net/http/httputil/#ReverseProxy

jtlisi:
@Dieterbe I named it gAuth because that was the alias of the package as it was currently being imported

cortex/cortex.go
Dieterbe: oh that's really nice

cortex/cortex.go
Dieterbe: ok

jtlisi:
@woodsaj Makes sense, this should be good. I implemented it slightly differently. I used one auth that just handles `"api_key`" as the username differently. I also made the file backend use a comma separated list of instance ID's associated with an org/password. That way one key can access all the instances of an org.

api/middleware.go
```diff
@@ -84,14 +84,41 @@ func (a *Api) Auth() macaron.Handler {
...
 				}
 			}
 		}
 		ctx.User = user
 	}
 }
 
+func (a *Api) CortexAuth() macaron.Handler {
+	return func(ctx *Context) {
+		instanceID, key, err := getInstanceIdAndApiKey(ctx)
```
woodsaj: lets get rid of our own getApiKey/getInstanceAndApiKey functions.  All they are doing is getting the user/password from an Authorization header.   the http.Request object has a function for that.
```
user, pass, ok := ctx.Req.BasicAuth()
```

auth/auth.go
```diff
@@ -23,12 +25,13 @@ func init() {
...
 type User struct {
-	OrgId   int
+	ID      int
 	IsAdmin bool
 }
 
+// AuthPlugins returns whether a api_key is a valid and if the user has access to a certain instance
 type AuthPlugin interface {
-	Auth(userKey string) (*User, error)
+	Auth(instanceID, password string) (*User, error)
```
woodsaj: lets use `username` instead of instanceId.  
The fact that the username might be an instanceId is an implementation detail that wont always be true for all plugins

api/middleware.go
```diff
@@ -84,14 +84,41 @@ func (a *Api) Auth() macaron.Handler {
...
+				ctx.JSON(401, err.Error())
+				return
+			}
+			log.Error(3, "failed to perform authentication: %q", err.Error())
+			ctx.JSON(500, err.Error())
+			return
+		}
+
+		ctx.User = user
+		ctx.Req.Request.Header.Add("X-Scope-OrgID", instanceID)
```
woodsaj: this needs to be `ctx.Req.Request.Header.Set()` so that any X-Scope-OrgId header set by the user is replaced.  If there was an existing header, Header.Add() would just append to it.

api/middleware.go
jtlisi: What about in cases where Bearer auth is used?

api/middleware.go
```diff
@@ -84,14 +84,41 @@ func (a *Api) Auth() macaron.Handler {
...
+	}
+}
+
 func getApiKey(c *Context) (string, error) {
```
woodsaj: lets get rid of this.  We should just call ctx.Req.Request.BasicAuth() instead.
The check to make sure the username == "apiKey" should be handled in the authPlugins.

auth/file.go
```diff
@@ -14,18 +16,21 @@ the user details associated with that key.
 
 example:
 ------------------
-[ahLaibi7Ee]
+[org1]
```
woodsaj: Sorry, i should not have told you to change this. I just realized we have some customers that send metrics via direct carbon and that uses this auth file.  To prevent a messy upgrade process, lets keep this file format compatible with how it was.

auth/gcom/auth.go
```diff
@@ -36,13 +37,12 @@ func (i *int64SliceFlag) String() string {
...
-	validTTL     = time.Minute * 5
-	invalidTTL   = time.Second * 30
-	authEndpoint = "https://grafana.net"
-	validOrgIds  = int64SliceFlag{}
-	cache        *AuthCache
-
-	Debug = false
+	validTTL      = time.Minute * 5
+	invalidTTL    = time.Second * 30
+	authEndpoint  = "https://grafana.net"
```
woodsaj: this should default to https://grafana.com

auth/auth.go
jtlisi: :+1:

api/middleware.go
jtlisi: :+1:

auth/file.go
jtlisi: :+1: easy fix

api/middleware.go
woodsaj: ok, lets just use a simple `getAuthCreds(ctx)` function that
- calls ctx.Req.Request.BasicAuth().  If basicAuth is used it should just return the user password
- else, check if there is a bearer token. if so, set the user to `api_key` and return the token.

auth/file.go
jtlisi: Actually, one issue with using the section header as the password is INI header sections decode as all lower case (at least with library we use) so they can't be used to store passwords that are cases sensitive.

woodsaj: That is fine for now.  This is basically just a hack to support direct carbon input anyway

jtlisi: https://github.com/go-ini/ini this will do the trick

jtlisi:
@woodsaj @Dieterbe Hey this is ready for another look

api/api.go
```diff
@@ -89,18 +90,20 @@ func (a *Api) Stop() {
...
 	m.Any("/graphite/*", a.Auth(), GraphiteProxy)
-	if *prometheusWriteEnabled {
-		m.Any("/prometheus/write", a.Auth(), PrometheusWrite)
+	if *prometheusMTWriteEnabled {
+		m.Any("/prometheus/write", a.Auth(), PrometheusMTWrite)
 	}
 	m.Any("/prometheus/*", a.Auth(), PrometheusProxy)
 	m.Post("/opentsdb/api/put", a.Auth(), OpenTSDBWrite)
+	m.Any("/api/prom/push", a.CortexAuth(), CortexWrite)
+	m.Any("/api/prom/*", a.CortexAuth(), CortexProxy)
```
Dieterbe: I thought that whether we write to cortex or to MT, we would use the same URL?
otherwise does this mean users have to reconfigure their clients if/when we decide to switch between cortex and MT?

(note: i'm not well versed here so i might be completely wrong)

auth/auth.go
```diff
@@ -23,12 +25,13 @@ func init() {
 }
 
 type User struct {
-	OrgId   int
+	ID      int
 	IsAdmin bool
 }
 
+// AuthPlugins returns whether a api_key is a valid and if the user has access to a certain instance
```
Dieterbe: an interface can't return anything. maybe you meant to put this with the Auth method?

auth/file.go
```diff
@@ -14,18 +14,20 @@ the user details associated with that key.
 
 example:
 ------------------
-[ahLaibi7Ee]
+[aaeipgnq]
 orgId = 1
 isAdmin = true
 
-[ubi5ZahD6l]
+[wpirgn123]
```
Dieterbe: why are these changing? it's especially confusing since the commit message says "return to original ini file auth" ?

Dieterbe: is there any significance to these examples or are they just random?

jtlisi: random, I just didn't want to go through the git log to get a random string lol

auth/auth.go
jtlisi: This refers to the auth method. I'll add a commit that clarifies

auth/grafana-com.go
```diff
@@ -11,25 +12,34 @@ func NewGrafanaComAuth() *GrafanaComAuth {
...
 		}
 		if err == gcom.ErrInvalidOrgId {
 			return nil, ErrInvalidOrgId
 		}
 		return nil, err
 	}
-	if instanceID != "api_key" {
-		err = u.CheckInstance(instanceID)
+
+	ctxUser := &User{
```
jtlisi: @woodsaj doesn't this lead open the possibility that if someone uses api_key as the username in basic auth they will send cortex data to whatever instance shares an id with their organization?

woodsaj: yep, that is a good point.  Pushing a new commit now to make the instance auth a separate auth plugin. This will be a much safer approach.

auth/auth.go
```diff
@@ -29,9 +29,9 @@ type User struct {
 	IsAdmin bool
 }
 
-// AuthPlugins returns whether a api_key is a valid and if the user has access to a certain instance
+// AuthPlugin is used to validate access
 type AuthPlugin interface {
-	Auth(username, password string) (*User, error)
+	Auth(username, password string) (*User, error) // Auth returns whether a api_key is a valid and if the user has access to a certain instance
```
Dieterbe: comments in go precede the declaration. see  https://blog.golang.org/godoc-documenting-go-code
that makes godoc work

jtlisi:
@woodsaj LGTM

We just need to make a note when setting up cortex instances to use the correct Auth plugin. I'll make sure it's noted in an issue in HM-API

PR #78, opened 2018-03-28
fix deadlock when queue.buffering.max.messages is reached

### LINKED ISSUE #77: Local: Queue full metrics
If you enable:

` config.SetKey("statistics.interval.ms", "5000")`
and setup something to log the metrics you can view the blocking queue

In my case i adjusted the max.messages to allow 1000000 vs 100000
`config.SetKey("queue.buffering.max.messages", "1000000")`

```
 "msg_cnt":**182288**
 "msg_max":1000000
```

82288 over the normal queue capacity.

The problematic part of this queue is that if you ever reach it tsdb-gw stops publishing to kafka and never recovers. Unfortunately it also still reports as healthy.
screen shot 2018-03-26 at 5 32 06 pm

I threw these into the logger on locally published build to look but this may be valuable to other users to have in the tsdb-gw graphed statistics.

woodsaj:
fixes #77

### DIFF
--- a/metric_publish/publish.go
+++ b/metric_publish/publish.go
@@ -143,7 +143,7 @@ func Publish(metrics []*schema.MetricData) error {
 
 	payload := make([]*kafka.Message, len(metrics))
 	pre := time.Now()
-	deliveryChan := make(chan kafka.Event)
+	deliveryChan := make(chan kafka.Event, len(metrics))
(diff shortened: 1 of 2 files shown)


woodsaj:
not sure how that line got removed. i added it back.

PR #79, opened 2018-03-28
bug fixes in instanceAuth

woodsaj:
- update URL to try and fetch specific hosted-metrics instance by id
- fix bugs that caused successfull auth for invalid instance ids.
- add unit tests for instance auth

### DIFF
--- a/auth/gcom/auth.go
+++ b/auth/gcom/auth.go
@@ -226,43 +226,36 @@ func Auth(adminKey, keyString string) (*SignedInUser, error) {
 	}
 
 	// add the user to the cache.
-
 	log.Debug("Auth: Caching validKey response for %d seconds", validTTL/time.Second)
-
 	cache.Set(keyString, user, validTTL)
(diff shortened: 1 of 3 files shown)


auth/gcom/auth.go
```diff
@@ -226,43 +226,36 @@ func Auth(adminKey, keyString string) (*SignedInUser, error) {
...
 
 		log.Debug("Auth: invalid instance key cached")
-
 		return ErrInvalidInstanceID
 	}
-	payload := url.Values{}
-	payload.Add("id", instanceID)
 
-	req, err := http.NewRequest("GET", fmt.Sprintf("%s/api/hosted-metrics", authEndpoint), nil)
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/api/hosted-metrics/%s", authEndpoint, instanceID), nil)
```
jtlisi: This is much better than getting a list of instances. :+1:

PR #80, opened 2018-03-28
[WIP] Tsdb gw refactor

jtlisi:
This is a starting point for the Refactor into two separate binaries.

Interested in suggestions on how publishing can work with both cortex and metrictank.
I'll comment specific lines of code with particulars


publish/publish.go
```diff
@@ -0,0 +1,39 @@
+package publish
+
+import (
+	log "github.com/sirupsen/logrus"
+	schema "gopkg.in/raintank/schema.v1"
+)
+
+type Publisher interface {
```
jtlisi: Created a publishing interface, that way the same API inputs can be used for a future Cortex write publisher. Ideally it would just replace `.` with `_`  and forward the request to cortex using the prometheus remote write format. All that would need to be done is write a publisher that takes our schema format and translates it into the prometheus write format. @woodsaj is this worth doing or would we be better off having entirely separate inputs or have an optimized format the exists only during translation in TSDB-GW?

util/log.go
```diff
@@ -0,0 +1,28 @@
+package util
```
jtlisi: Adding a simple log config unit that can be used to set the log level. Uses logrus. Easy to add line/filenumber to this package similar to Prometheus/common/log, we can just only add it for WARN and above

util/tracer.go
```diff
@@ -4,13 +4,13 @@ import (
...
 	opentracing "github.com/opentracing/opentracing-go"
-	"github.com/raintank/worldping-api/pkg/log"
+	log "github.com/sirupsen/logrus"
 	jaeger "github.com/uber/jaeger-client-go"
 	jaegercfg "github.com/uber/jaeger-client-go/config"
 	jaegerlog "github.com/uber/jaeger-client-go/log"
 )
 
-func GetTracer(enabled bool, addr string) (opentracing.Tracer, io.Closer, error) {
+func GetTracer(app string, enabled bool, addr string) (opentracing.Tracer, io.Closer, error) {
```
jtlisi: Making the tracer have a configurable name span

metrictank/ingest/ingest.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+var (
+	metricPool    = metricpool.NewMetricDataPool()
+	schemas       *conf.Schemas
+	schemaFile    = flag.String("api-schemas-file", "/etc/storage-schemas.conf", "path to carbon storage-schemas.conf file")
+	ingestEnabled = flag.Bool("api-ingest", true, "enable api ingest for datadog/opentsdb/prometheus for metrictank")
+)
+
+func IngestInit() {
```
jtlisi: Share the schema matching file for all API based ingest files. Ideally carbon can share this as well, or is there any good reason to separate the two?

jtlisi:
Replacing the worldping logs is gonna require handpicking through the current log lines and reformatting them, since some are gonna look weird due to function changes

publish/publish.go
woodsaj: I think this approach will be just fine.  The only real concern is the extra performance overhead of marshalling from `input format` -> `schema.MetricData` -> `prometheus`.    But a small performance penalty for simpler code is fine.

util/log.go
woodsaj: yes, lets add the file/linenumbers for warn and error messages.

api/middleware.go
```diff
@@ -179,7 +183,7 @@ func pathSlug(p string) string {
 	return strings.Replace(strings.Replace(slug, "/", "_", -1), ".", "_", -1)
 }
 
-func Tracer() macaron.Handler {
+func Tracer(spanName string) macaron.Handler {
```
Dieterbe: this string is used to set the component and seems to have nothing to do with spans.
it needs a better name

publish/publish.go
jtlisi: @woodsaj Agreed, especially since most cortex metrics won't go through this pathway, they will just be proxied to cortex

api/middleware.go
jtlisi: :+1:

jtlisi:
@Dieterbe @tomwilkie @woodsaj @replay I think this is ready for another look.

cmd/cortex-gw/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# copy the built binaries from the build image
+COPY --from=build $BASE_PATH/build/cortex-gw /usr/bin/cortex-gw
+
+EXPOSE 80
+EXPOSE 443
+EXPOSE 8001
+
+ENTRYPOINT ["/usr/bin/entrypoint.sh"]
+CMD ["/usr/bin/cortex-gw", "-config=/etc/raintank/tsdb.ini"]
```
Dieterbe: it's overdue time to get rid of the raintank name.
if the config is just 1 file and we're confident it'll remain that way, why not just use `/etc/cortex-gw.ini` ? or if we need multiple files or we think that's likely in the future, how about `/etc/cortex-gw/gw.ini` ?

cmd/tsdb-gw/main.go
```diff
@@ -105,39 +91,42 @@ func main() {
...
 
-	metric_publish.Init(*broker)
-
-	if err := graphite.Init(*graphiteUrl); err != nil {
-		log.Fatal(4, err.Error())
+	publisher := kafka.New(*broker)
+	if publisher == nil {
+		publish.Init(nil)
+	} else {
+		publish.Init(publisher)
```
Dieterbe: seems like you can just `publish.Init(publisher)` without doing the if/else branches

cmd/cortex-gw/main.go
```diff
@@ -0,0 +1,134 @@
...
+	log.Infof("Starting %v ...", app)
+	done := make(chan struct{})
+	inputs = append(inputs, api.Start(), ms)
+	go handleShutdown(done, interrupt, inputs)
+	log.Infof("%v Started", app)
+	<-done
+}
+
+// Stoppable represents things that can be stopped.
+type Stoppable interface {
```
Dieterbe: this and the handleShutdown is neat.

cmd/tsdb-gw/main.go
```diff
@@ -105,39 +91,42 @@ func main() {
...
 
+	log.Infof("Starting %v ...", app)
 	done := make(chan struct{})
-	inputs = append(inputs, api.InitApi(), carbon.InitCarbon(), ms)
+	inputs = append(inputs, api.Start(), carbon.InitCarbon())
 	go handleShutdown(done, interrupt, inputs)
-
+	log.Infof("%v Started", app)
 	<-done
 }
```
Dieterbe: line below: is it deliberate to instantiate a prometheus /metrics endpoint in addition to the graphite stats sender?

cmd/cortex-gw/Dockerfile
jtlisi: Agreed, might as well do it now

cmd/tsdb-gw/main.go
jtlisi: @Dieterbe I thought that but when I tried it I encountered some unexpected behavior. I can take another look and see what was going on if you want. But it seemed like even if publisher was nil, because it was typed it would cause `publish.Init(publisher)` to have unexpected behavior

Dieterbe: ah probably because it's a non-nil interface with a nil value embedded in it

api/api.go
```diff
@@ -9,70 +9,74 @@ import (
...
+		log.Fatal(err.Error())
 	}
 	a.l = l
-
-	PrometheusMTInit()
+	m := macaron.New()
+	m.Use(macaron.Recovery())
+	m.Use(macaron.Renderer())
+	m.Use(Tracer(appName))
 	a.InitRoutes(m)
```
woodsaj: I think we can just move InitRoutes() into here.

cmd/cortex-gw/Dockerfile
woodsaj: we should also move this entire repo to the grafana org

cmd/cortex-gw/main.go
```diff
@@ -0,0 +1,134 @@
...
+	log.Infoln("shutdown started.")
+	var wg sync.WaitGroup
+	for _, input := range inputs {
+		wg.Add(1)
+		go func(plugin Stoppable) {
+			plugin.Stop()
+			wg.Done()
+		}(input)
+	}
+	wg.Wait()
```
woodsaj: If any of the plugins block and dont return, the app will never shutdown.  I think we should have a max wait time here.
```
complete := make(chan struct{})
go func() {
   wg.Wait()
   close(complete)
}()
timer := time.NewTimer(time.Minute)
select {
case <-timer.C:
 log.Errorln("shutdown taking too long, giving up waiting on plugins")
case <-complete:
  log.Infof("shutdown complete")
}
```

ingest/datadog.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/raintank/tsdb-gw/api"
+	"github.com/raintank/tsdb-gw/publish"
+	log "github.com/sirupsen/logrus"
+	schema "gopkg.in/raintank/schema.v1"
+)
+
+// DataDogPayload struct to unmarshal datadog agent json
+type DataDogPayload struct {
+	Series []struct {
```
woodsaj: This is missing the MType
it is an int, where 
0=gauge
1=rate
2-count

ingest/datadog.go
```diff
@@ -0,0 +1,112 @@
...
+			for _, point := range ts.Points {
+				md := metricPool.Get()
+				*md = schema.MetricData{
+					Name:     ts.Name,
+					Metric:   ts.Name,
+					Interval: s.Retentions[0].SecondsPerPoint,
+					Value:    point.Value,
+					Unit:     "unknown",
+					Time:     int64(point.Ts),
+					Mtype:    "gauge",
```
woodsaj: this should use the Mtype sent by the dd-agent

cmd/tsdb-gw/main.go
woodsaj: I dont think we should use prometheus metrics for tsdb-gw yet.  Doesnt look like we are actually collecting any prometheus metrics anyway. So lets get rid of this code.

metrictank/metrics.go
```diff
@@ -1,4 +1,4 @@
-package api
+package metrictank
```
woodsaj: Shouldnt this all be in the "ingest" package?

cmd/tsdb-gw/main.go
gouthamve: As an aside, I am not sure if we are exporting `go_` stats to graphite, like `goroutines`, `memory`, `gc` and others? If not, they sometimes help in debugging and we could enable them in Graphite, or we expose them via prometheus and make `-gw` a target to Prometheus in k8s?

See: https://pastebin.com/KFb5Lc24 for the stats exposed.

woodsaj:
Overall this is looking really good.  A few changes needed, but nothing major.

api/middleware.go
```diff
@@ -84,8 +84,12 @@ func (a *Api) Auth() macaron.Handler {
 		}
 
 		if key == "" {
-			ctx.JSON(401, "Unauthorized")
-			return
+			key = ctx.Req.Header.Get("Dd-Api-Key")
```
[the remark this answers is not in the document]
jtlisi: But we still need to parse the "dd-api-key" header? Or else how will we get the api-key?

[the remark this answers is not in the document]
woodsaj: we can either
a) just not use the dd-api-key and instead include the user/password in the app_url setting in the dd-agent
b) the dd-agent ingest handler can use its own auth middleware

api/api.go
```diff
@@ -9,70 +9,82 @@ import (
...
 
 type Api struct {
 	l          net.Listener
 	done       chan struct{}
 	authPlugin auth.AuthPlugin
+	Router     *macaron.Macaron
 }
 
-func InitApi() *Api {
+type ApiConfig struct {
```
woodsaj: This doesnt look to be use anywhere

ingest/datadog.go
```diff
@@ -0,0 +1,113 @@
...
+type DataDogPayload struct {
+	Series []struct {
+		Name   string `json:"metric"`
+		Points []struct {
+			Ts    float64
+			Value float64
+		} `json:"points"`
+		Tags   []string `json:"tags"`
+		Host   string   `json:"host"`
+		Mtype  string   `json:"types"`
```
woodsaj: MType is an int, and the json attribute is `type`
0=gauge
1=rate
2-count

jtlisi: When I run the agent locally it seems to marshal the json with the Mtype as a string?

api/api.go
jtlisi: Sorry about that, I was tinkering with config structs but decided it would be better to add it in a separate PR

ingest/datadog.go
jtlisi:
```json
{
        "metric": "system.uptime",
        "points": [
                [
                        1522874984,
                        20680
                ]
        ],
        "tags": null,
        "host": "jtl-matebook",
        "type": "gauge",
        "interval": 0,
        "source_type_name": "System"
}
```

README.md
```diff
@@ -0,0 +1,21 @@
+# Grafana Gateways
+
+Ingest and query metrics using [grafana.com](www.grafana.com) accounts/apikeys as authentication. Handles traffic toward [Cortex](https://github.com/weaveworks/cortex) and [Metrictank](https://github.com/grafana/metrictank).
```
replay: since `www.grafana.com` is redirecting to `grafana.com`, i think just for tidiness this should better refer to `grafana.com` without `www.`

ingest/datadog.go
```diff
@@ -0,0 +1,110 @@
...
+func DataDogMTWrite(ctx *api.Context) {
+	if ctx.Req.Request.Body != nil {
+		defer ctx.Req.Request.Body.Close()
+
+		var body []byte
+		var err error
+		if ctx.Req.Request.Header.Get("Content-Encoding") == "deflate" {
+			zr, err := zlib.NewReader(ctx.Req.Request.Body)
+			if err != nil {
+				panic(err)
```
replay: Won't panic stop the whole process? Wouldn't it be more appropriate to write an error and return?

jtlisi: Yes, good catch. A panic would be terrible here

woodsaj: this is a HTTP Handler, the panic will be caught by macaron and a stack strace logged. Which is probably good enough

woodsaj: But if the headers imply that the content is compressed, and we cant decompress it, that is a 400 error (bad user input).

ingest/datadog.go
```diff
@@ -0,0 +1,110 @@
...
+
+		var series DataDogPayload
+		err = json.Unmarshal(body, &series)
+		if err != nil {
+			return
+		}
+
+		buf := make([]*schema.MetricData, 0)
+		for _, ts := range series.Series {
+			_, s := schemas.Match(ts.Name, 0)
```
replay: should we now support schema patterns that include tags?
if so, then this should probably be something like:
```
tagSet := createTagSet(ts.Host, ts.Device, ts.Tags)
_, s := schemas.Match(ts.Name + ";" + tagSet, 0)
```

ingest/datadog.go
```diff
@@ -0,0 +1,110 @@
...
+					Tags:     tagSet,
+					OrgId:    ctx.ID,
+				}
+				md.SetId()
+				buf = append(buf, md)
+			}
+		}
+
+		err = publish.Publish(buf)
+		for _, m := range buf {
```
replay: i think putting them back into the pool could be done async after returning

ingest/datadog.go
```diff
@@ -0,0 +1,110 @@
...
+		Points [][2]float64 `json:"points"`
+		Tags   []string     `json:"tags"`
+		Host   string       `json:"host"`
+		Mtype  string       `json:"type"`
+		Device string       `json:"device,omitempty"`
+	} `json:"series"`
+}
+
+func DataDogMTWrite(ctx *api.Context) {
+	if ctx.Req.Request.Body != nil {
```
replay: it would safe some indentation in the whole function to just return if `.Body == nil`.

scripts/build.sh
```diff
@@ -19,19 +19,19 @@ fi
...
 # Build binary
-go build -tags static -ldflags "-X main.GitHash=$GIT_VERSION" -o $BUILD_DIR/tsdb-gw
-cd cmd/tsdb-usage
-go build -tags static -ldflags "-X main.GitHash=$GIT_VERSION" -o $BUILD_DIR/tsdb-usage
+cd cmd/tsdb-gw
+CGO_ENABLED=1 go build -tags static -ldflags "-X main.GitHash=$GIT_VERSION" -o $BUILD_DIR/tsdb-gw
+cd ../tsdb-usage
+CGO_ENABLED=1 go build -tags static -ldflags "-X main.GitHash=$GIT_VERSION" -o $BUILD_DIR/tsdb-usage
+cd ../cortex-gw
+go build -tags static -ldflags "-X main.GitHash=$GIT_VERSION" -o $BUILD_DIR/cortex-gw
```
replay: i think this shouldn't need the `-tags static`

api/middleware.go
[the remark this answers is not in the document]
jtlisi: @woodsaj Just clarifying the first option is set for now. We can always add a second auth for the agent later.

PR #81, opened 2018-03-30
Add prometheus metrics for cortex-gw

gouthamve:
This is a quick stopgap as gw is moving to gorilla-mux and we'll be able to use other middlewares.

---
This change is [Reviewable](https://reviewable.io/reviews/raintank/tsdb-gw/81)


gouthamve:
Just a note that care has to be taken not to expose the `/metrics` endpoint.

api/middleware.go
```diff
@@ -14,12 +14,26 @@ import (
...
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/macaron.v1"
 )
 
+var (
+	requestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "tsdb_gw",
+		Name:      "request_duration_seconds",
+		Help:      "Time (in seconds) spent serving HTTP requests.",
+		Buckets:   []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10, 25, 50, 100},
```
tomwilkie: Thats a lot of buckets - generally if its anything other that ExponentialBuckets it needs a comment explaining why.  And less than 10 buckets pls.

tomwilkie:
Just the one comment, otherwise LGTM!

api/middleware.go
gouthamve: Hmm, so my understanding is that after this moves to `mux`, we'll be using the ww middleware anyways for instrumentation.

So I took it from there: https://github.com/weaveworks/common/blob/a2a9c9a0d6d4be4de3bf0400e200ecb15e60de7c/instrument/instrument.go#L14-L16

jtlisi:
LGTM

PR #83, opened 2018-03-31
Add prometheus metrics to cortex endpoints

gouthamve:
This is basically the last two PRs but onto master.

The prometheus metrics server is always on, I can disable it with another flag if required.

---
This change is [Reviewable](https://reviewable.io/reviews/raintank/tsdb-gw/83)


main.go
```diff
@@ -45,6 +47,8 @@ var (
 
 	tracingEnabled = flag.Bool("tracing-enabled", false, "enable/disable distributed opentracing via jaeger")
 	tracingAddr    = flag.String("tracing-addr", "localhost:6831", "address of the jaeger agent to send data to")
+
+	metricsAddr = flag.String("metrics-addr", "localhost:8001", "http service address for the /metrics endpoint")
```
tomwilkie: Sensible defaults: no point in making it only listen on `localhost` as we'll be deploying this on k8s.

PR #85, opened 2018-04-07
Listen for metrics on all interfaces by default

gouthamve:
This will fix the broken scrapes.

---
This change is [Reviewable](https://reviewable.io/reviews/raintank/tsdb-gw/85)


jtlisi:
@gouthamve LGTM

PR #86, opened 2018-04-07
add debug logs to gcom instance auth

jtlisi:


jtlisi:
@gouthamve just adding some debug logs to see what is going on with the gw

PR #87, opened 2018-04-08
update config file paths

Dieterbe:
* fix config paths, naming config file after the commands/binaries.
  this solves:
  * two of these services use tsdb.ini, yet they are not TSDB's
  * one uses gw.ini yet we have 2 different gw services
  * tsdb-usage was still using raintank path
* update upstart and systemd configs that still used pre-refactor raintank paths
* remove ambiguity: scripts/config/tsdb.ini should be
  scripts/config/tsdb-gw.ini ; note that
  we don't have a default config file for cortex-gw yet,
  previously it would errorously add the tsdb-gw's config file.
  scripts/config/cortex-gw.ini still has to be created.
* fix all pre-refactor raintank paths from build scripts

* update tsdb auth config path

@jtlisi thoughts?


jtlisi:
@Dieterbe This looks fine. It doesn't look like any changes are required to hosted-metrics-api either. Could you add the cortex-gw.ini file to the PR so circleCI passes then I can merge

briangann:
Since the names of the config files are changing, it would be useful to also relocate them into /etc/raintank vs living inside /etc.

This makes overlaying configmaps inside k8 trivial, and also overlays using docker-compose much easier (volumes with hostpaths etc).

The path would be:
/etc/raintank/tsdb-gw.ini vs /etc/tsdb-gw.ini

and the startup would just use -config /etc/raintank/tsdb-gw.ini 

also - the example .ini references /etc/raintank/tsdb-auth.ini, which would end up placing all of the configuration files in the same location.

Dieterbe:
> dieter [6:33 PM]
so you're saying loading single files via the volume mechanism of docker or configmap mechanism of k8s is trickier/harder than doing directories?
> brian.gann [6:33 PM]
yeah its buggy if anything
lots of bug reports in k8/docker
using subPath doesn’t work on configMaps
so it’s way easier to just have a configmap that has a bunch of files that get mounted into a directory

so I think:
1) we should use directories again
2) but not a raintank directory. we're trying to move away from that name.

I suggest: /etc/gw/
within that directory then goes tsdb-gw.ini, cortex-gw, etc.

PR #88, opened 2018-04-10
cortex publish interface

jtlisi:
this adds a publisher pipeline for cortex so datadog/opentsdb/etc can be written to cortex.

#### Notes
* `.` characters are replaced wholesale on write with `_`
* The publish code could benefit from a metrics queue and some rate limiting features. I'd rather keep it simple at the moment and add performance optimizing features as needed.

### DIFF
--- a/cmd/cortex-gw/main.go
+++ b/cmd/cortex-gw/main.go
@@ -17,7 +17,7 @@ import (
 	"github.com/raintank/tsdb-gw/ingest"
 	"github.com/raintank/tsdb-gw/ingest/carbon"
 	"github.com/raintank/tsdb-gw/publish"
-	cortex_publish "github.com/raintank/tsdb-gw/publish/cortex"
+	cortexPublish "github.com/raintank/tsdb-gw/publish/cortex"
 	"github.com/raintank/tsdb-gw/query/cortex"
 	"github.com/raintank/tsdb-gw/util"
 	log "github.com/sirupsen/logrus"
@@ -67,10 +67,10 @@ func main() {
 	}
 	defer traceCloser.Close()
 
-	inputs := make([]Stoppable, 0)
+	var inputs []Stoppable
 
 	if *cortexWriteEnabled {
-		publish.Init(cortex_publish.NewCortexPublisher())
+		publish.Init(cortexPublish.NewCortexPublisher())
 		inputs = append(inputs, carbon.InitCarbon())
 	} else {
 		publish.Init(nil)
@@ -131,7 +131,7 @@ func handleShutdown(done chan struct{}, interrupt chan os.Signal, inputs []Stopp
 
 // InitRoutes initializes the routes.
 func initRoutes(a *api.Api) {
-	a.Router.Any("/api/prom/push", a.PromStats("cortex-write"), a.Auth(), cortex_publish.Write)
+	a.Router.Any("/api/prom/push", a.PromStats("cortex-write"), a.Auth(), cortexPublish.Write)
 	a.Router.Any("/api/prom/*", a.PromStats("cortex-read"), a.Auth(), cortex.Proxy)
 	a.Router.Post("/datadog/api/v1/series", a.Auth(), ingest.DataDogWrite)
 	a.Router.Post("/opentsdb/api/put", a.Auth(), ingest.OpenTSDBWrite)
--- a/cmd/cortex-gw/main.go
+++ b/cmd/cortex-gw/main.go
@@ -24,12 +24,12 @@ import (
 )
 
 var (
-	app                = "cortex-gw"
-	GitHash            = "(none)"
-	showVersion        = flag.Bool("version", false, "print version string")
-	confFile           = flag.String("config", "/etc/gw/cortex-gw.ini", "configuration file path")
-	authPlugin         = flag.String("api-auth-plugin", "grafana-instance", "auth plugin to use. (grafana-instance|file)")
-	cortexWriteEnabled = flag.Bool("cortex-forward-3rdparty", false, "enable writing to cortex with non standard agents")
+	app             = "cortex-gw"
+	GitHash         = "(none)"
+	showVersion     = flag.Bool("version", false, "print version string")
+	confFile        = flag.String("config", "/etc/gw/cortex-gw.ini", "configuration file path")
+	authPlugin      = flag.String("api-auth-plugin", "grafana-instance", "auth plugin to use. (grafana-instance|file)")
+	forward3rdParty = flag.Bool("forward-3rdparty", false, "enable writing to cortex with non standard agents")
 
 	tracingEnabled = flag.Bool("tracing-enabled", false, "enable/disable distributed opentracing via jaeger")
 	tracingAddr    = flag.String("tracing-addr", "localhost:6831", "address of the jaeger agent to send data to")
@@ -69,7 +69,7 @@ func main() {
 
 	var inputs []Stoppable
 
-	if *cortexWriteEnabled {
+	if *forward3rdParty {
 		publish.Init(cortexPublish.NewCortexPublisher())
 		inputs = append(inputs, carbon.InitCarbon())
 	} else {
--- a/cmd/cortex-gw/main.go
+++ b/cmd/cortex-gw/main.go
@@ -29,7 +29,7 @@ var (
 	showVersion        = flag.Bool("version", false, "print version string")
 	confFile           = flag.String("config", "/etc/gw/cortex-gw.ini", "configuration file path")
 	authPlugin         = flag.String("api-auth-plugin", "grafana-instance", "auth plugin to use. (grafana-instance|file)")
-	cortexWriteEnabled = flag.Bool("cortex-publish-enabled", false, "enable writing to cortex with non standard agents")
+	cortexWriteEnabled = flag.Bool("cortex-forward-3rdparty", false, "enable writing to cortex with non standard agents")
 
 	tracingEnabled = flag.Bool("tracing-enabled", false, "enable/disable distributed opentracing via jaeger")
 	tracingAddr    = flag.String("tracing-addr", "localhost:6831", "address of the jaeger agent to send data to")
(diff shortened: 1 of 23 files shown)


publish/cortex/publish.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+type cortexPublisher struct {
+	url     *url.URL
+	client  *http.Client
+	timeout time.Duration
+}
+
+func NewCortexPublisher() *cortexPublisher {
+	CortexWriteURL, err := url.Parse(*cortexWriteURL)
```
woodsaj: this is already parsed in `init()`

cmd/cortex-gw/main.go
```diff
@@ -118,8 +131,10 @@ func handleShutdown(done chan struct{}, interrupt chan os.Signal, inputs []Stopp
 
 // InitRoutes initializes the routes.
 func initRoutes(a *api.Api) {
-	a.Router.Any("/api/prom/push", a.PromStats("cortex-write"), a.Auth(), cortex.Write)
+	a.Router.Any("/api/prom/push", a.PromStats("cortex-write"), a.Auth(), cortex_publish.Write)
 	a.Router.Any("/api/prom/*", a.PromStats("cortex-read"), a.Auth(), cortex.Proxy)
+	a.Router.Post("/datadog/api/v1/series", a.Auth(), ingest.DataDogMTWrite)
```
Dieterbe: MT write ?

publish/cortex/publish.go
```diff
@@ -0,0 +1,212 @@
...
+	requestDuration.WithLabelValues("succeeded").Observe(took.Seconds())
+	return nil
+}
+
+func (c *cortexPublisher) Type() string {
+	return "cortex"
+}
+
+// Store sends a batch of samples to the HTTP endpoint.
+func (c *cortexPublisher) Write(req *prompb.WriteRequest) error {
```
Dieterbe: Comment does not match method name

cmd/cortex-gw/main.go
```diff
@@ -118,8 +131,10 @@ func handleShutdown(done chan struct{}, interrupt chan os.Signal, inputs []Stopp
...
+	a.Router.Post("/opentsdb/api/put", a.Auth(), ingest.OpenTSDBWrite)
```
Dieterbe: Is there a reason why some write methods include the backend type and others don't ?

jtlisi:
@Dieterbe This should be better

publish/cortex/publish.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if httpResp.StatusCode/100 != 2 {
+		scanner := bufio.NewScanner(io.LimitReader(httpResp.Body, maxErrMsgLen))
+		line := ""
+		if scanner.Scan() {
+			line = scanner.Text()
+		}
+		err = fmt.Errorf("server returned HTTP status %s: %s", httpResp.Status, line)
+	}
+	if httpResp.StatusCode/100 == 5 {
```
replay: what's the point of this `if`? it seems like whether the condition evaluates to `true`or not the result is always `return err`

publish/cortex/publish.go
```diff
@@ -0,0 +1,226 @@
...
+	return err
+}
+
+type writeRequest struct {
+	Request *prompb.WriteRequest
+	orgID   int
+}
+
+func packageMetrics(metrics []*schema.MetricData) (*writeRequest, error) {
+	req := &prompb.WriteRequest{
```
replay: i'd move this instantiation below the `if len(metrics) < 1`. It's not necessary to do that if `len(metrics) == 0`

publish/cortex/publish.go
```diff
@@ -0,0 +1,226 @@
...
+		labels := make([]*prompb.Label, 0, len(m.Tags)+1)
+		labels = append(labels,
+			&prompb.Label{
+				Name:  "__name__",
+				Value: strings.Replace(m.Name, ".", "_", -1),
+			},
+		)
+		for _, tag := range m.Tags {
+			tv := strings.SplitN(tag, "=", 2)
+			if len(tv) < 2 || tv[1] == "" {
```
replay: should `tv[0] == ""` be valid in this case?

publish/cortex/publish.go
Dieterbe: Is it a common prometheus style or something, to instantiate variables as pointed-to addresses?
In contrast to the style of instantiating as an actual value, and later referencing it via `&` to signify we want to share it.
https://www.ardanlabs.com/blog/2017/05/language-mechanics-on-escape-analysis.html makes a nice case IMHO about why the 2nd style is better. I highly recommend reading this entire 4 part series.

jtlisi:
@replay This should do the trick

cmd/cortex-gw/main.go
```diff
@@ -14,17 +14,22 @@ import (
...
-	GitHash     = "(none)"
-	showVersion = flag.Bool("version", false, "print version string")
-	confFile    = flag.String("config", "/etc/gw/cortex-gw.ini", "configuration file path")
-	authPlugin  = flag.String("api-auth-plugin", "grafana-instance", "auth plugin to use. (grafana-instance|file)")
+	app                = "cortex-gw"
+	GitHash            = "(none)"
+	showVersion        = flag.Bool("version", false, "print version string")
+	confFile           = flag.String("config", "/etc/gw/cortex-gw.ini", "configuration file path")
+	authPlugin         = flag.String("api-auth-plugin", "grafana-instance", "auth plugin to use. (grafana-instance|file)")
+	cortexWriteEnabled = flag.Bool("cortex-publish-enabled", false, "enable writing to cortex with non standard agents")
```
Dieterbe: this is confusing because the help says "enable writing to cortex with non standard agents" but according to the source code it simply controls whether writing to cortex to is enabled or not, irrespective of which agents are used. and the variable name also implies this usage.  do i understand this correctly? furthermore, what else would we publish to, given that this is cortex-gw?
should this be simply:
```
write-enabled # enable writing to cortex
```
?

cmd/cortex-gw/main.go
```diff
@@ -14,17 +14,22 @@ import (
 	"github.com/grafana/globalconf"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/raintank/tsdb-gw/api"
+	"github.com/raintank/tsdb-gw/ingest"
+	"github.com/raintank/tsdb-gw/ingest/carbon"
+	"github.com/raintank/tsdb-gw/publish"
+	cortex_publish "github.com/raintank/tsdb-gw/publish/cortex"
```
Dieterbe: https://golang.org/doc/effective_go.html#mixed-caps

cmd/cortex-gw/main.go
```diff
@@ -62,6 +67,15 @@ func main() {
 	}
 	defer traceCloser.Close()
 
+	inputs := make([]Stoppable, 0)
```
Dieterbe: `var inputs []Stoppable`.
https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful

publish/cortex/publish.go
```diff
@@ -0,0 +1,226 @@
...
+	httpReq.Header.Set("X-Prometheus-Remote-Write-Version", "0.1.0")
+	httpReq.Header.Set("X-Scope-OrgID", strconv.Itoa(req.orgID))
+
+	ctx, cancel := context.WithTimeout(context.Background(), c.timeout)
+	defer cancel()
+
+	httpResp, err := ctxhttp.Do(ctx, c.client, httpReq)
+	if err != nil {
+		// Errors from client.Do are from (for example) network errors, so are
+		// recoverable.
```
Dieterbe: what's the significance of recoverable vs non-recoverable? and why does it matter, given that 15 lines higher the recoverability is opposite but the outcome is the same (returning the error to caller)

publish/cortex/publish.go
```diff
@@ -0,0 +1,226 @@
...
+			line = scanner.Text()
+		}
+		err = fmt.Errorf("server returned HTTP status %s: %s", httpResp.Status, line)
+	}
+
+	return err
+}
+
+type writeRequest struct {
+	Request *prompb.WriteRequest
```
Dieterbe: do we need a pointer to it? considering it's just a slice?

cmd/cortex-gw/main.go
jtlisi: Prometheus remote write requests to cortex are forwarded to cortex despite whether this is enabled or not. This allows for other metrics to be packaged into the prometheus remote write format and forwarded to cortex.

publish/cortex/publish_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/prometheus/prometheus/prompb"
+	schema "gopkg.in/raintank/schema.v1"
+)
+
+func Test_packageMetrics(t *testing.T) {
+	tests := []struct {
+		name    string
+		metrics []*schema.MetricData
+		want    *writeRequest
```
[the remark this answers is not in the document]
jtlisi: I agree, I'll refactor to use less pointers

cmd/cortex-gw/main.go
jtlisi: :+1:

cmd/cortex-gw/main.go
Dieterbe: OK then this option should change the name, because it is too easy to mistake it for an option that controls whether writes to cortex are enabled or not (in general)

maybe.. `forward-3rdparty` or something?

publish/kafka/publish.go
```diff
@@ -153,6 +167,10 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
 	partitioner := partitionerPool.Get().(Partitioner)
 
 	for i, metric := range metrics {
+		_, s := m.schemas.Match(metric.Name, 0)
+		metric.Interval = s.Retentions[0].SecondsPerPoint
+		metric.SetId()
+
```
Dieterbe: I like your idea of moving the schema matching into the publisher, rather than the input.
but note that some inputs may have already set the interval (e.g. clients sending MetricData payloads to metrictank)
so we only need to match if `.Interval == 0`

PR #89, opened 2018-04-11
Optimize kafka format

Dieterbe:
counterpart to https://github.com/grafana/metrictank/pull/876


publish/kafka/publish.go
```diff
@@ -153,10 +168,37 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
...
+			ok := keyCache.Touch(mkey, pre)
+			// we've seen this key recently. we can use the optimized format
+			if ok {
+				data = bufferPool33.Get()
+				mp := schema.MetricPoint{
+					MKey:  mkey,
+					Value: metric.Value,
+					Time:  uint32(metric.Time),
+				}
+				data[:1][0] = byte(msg.FormatMetricPoint)
```
woodsaj: should this be `data[0]`

Dieterbe: no. i realize this looks a little funky, but data has len 0 and cap 33 so this is just a way to ensure the first byte is set correctly

woodsaj: i have never seem this before.  But it doesnt appear to work for me.
https://play.golang.org/p/ZvxgxydI5YU

Dieterbe: Yeah it looks a bit unusual for sure. didn't see a cleaner way to do it though.
some backstory:
when coming up with the marshaling and unmarshaling for MetricPoint, I took inspiration from how `msgp` does it. The msgp marshaling code encodes data in the space between length and capacity, as opposed to between 0 and length, so followed that pattern.
By extension, the format header is also stored in the first byte after the length (which we know to be zero for byte slices coming out of the pool) which is also why we pass `data[1:]` to the marshaling method: to have it put the binary data after the first byte (which we want to keep).

there is a bug in my code though, which is that after the Marshal call, we should set `data = data[:33]`
that also makes your example work: https://play.golang.org/p/eS7XQdZ8XpW (imagine that the marshal code used bytes 0 through 9 to store its data)

Dieterbe:
regarding the sync.Map, https://golang.org/pkg/sync/#Map reads pretty discouraging:

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination.

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow

does not apply to us I think. almost every operation is an update and we also prune regularly.

> or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.

also doesn't apply.

furthermore, sync.Map has downside of no type safety, lots of casting, and using interfaces, which I think would allocate.

sharded maps are commonly recommended on the performance channel of gophers slack for this sort of stuff, and I think it makes sense. I agree though that the keycache stuff is borderline too complicated, though I would say it's mainly caused by the Stamp and Ref types, and in a lesser amount due to the sharded and mutex-protected maps. perhaps the Len methods can be dropped, they don't seem to be used.

publish/kafka/keycache/cache.go
```diff
@@ -0,0 +1,44 @@
...
+func NewCache(ref Ref) *Cache {
+	c := Cache{}
+	for i := 0; i < 256; i++ {
+		c.shards[i] = NewShard(ref)
+	}
+	return &c
+}
+
+func (c *Cache) Touch(key schema.Key, t time.Time) bool {
+	shard := int(key[0])
```
woodsaj: can you add a comment for why key[0] is used as the shard key.
ie, key is a [16]byte hash of the metric attributes.  So the first byte will be evenly distributed. (i actually just scanned all 32million keys from one of our deployments to confirm that the first byte is evenly distributed)

publish/kafka/keycache/keycache.go
```diff
@@ -0,0 +1,90 @@
...
+		staleThresh:   uint8(staleThresh.Nanoseconds() / 1e9 / 600),
+		caches:        make(map[uint32]*Cache),
+	}
+	go k.prune()
+	return k
+}
+
+// marks the key as seen and returns whether it was seen before
+func (k *KeyCache) Touch(key schema.MKey, t time.Time) bool {
+	k.Lock()
```
woodsaj: This seems like a bad design.   Why have all the effort of making a concurrent cache, and they require all rights to be gated by a single lock.

publish/kafka/keycache/keycache.go
```diff
@@ -0,0 +1,90 @@
...
+	schema "gopkg.in/raintank/schema.v1"
+)
+
+type SubKey [15]byte
+
+type KeyCache struct {
+	staleThresh   uint8 // number of 10-minutely periods
+	pruneInterval time.Duration
+
+	sync.Mutex
```
replay: wouldn't an RWMutex make sense here?

publish/kafka/keycache/shard.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func NewShard(ref Ref) Shard {
+	return Shard{
+		ref: ref,
+		m:   make(map[SubKey]Stamp),
+	}
+}
+
+// callers responsability that t >= ref and t-ref <= 42 hours
```
replay: `responsibility`

publish/kafka/keycache/shard.go
```diff
@@ -0,0 +1,104 @@
...
+func (s *Shard) Touch(key schema.Key, t time.Time) bool {
+	var sub SubKey
+	copy(sub[:], key[1:])
```
replay: i'm not 100% sure, but i think this copying is only necessary if `s.m[sub]` does not exist and needs to be created, right? but for the check whether it exists or not `s.m[key[1:]]` would be sufficient and save some copying

publish/kafka/keycache/shard.go
```diff
@@ -0,0 +1,104 @@
...
+	s.Lock()
+	_, ok := s.m[sub]
+	s.m[sub] = NewStamp(s.ref, t)
+	s.Unlock()
+	return ok
+}
+
+func (s *Shard) Len() int {
+	s.Lock()
```
replay: also only needs read lock

publish/kafka/keycache/shard.go
```diff
@@ -0,0 +1,104 @@
...
+
+func NewRef(t time.Time) Ref {
+	unix := t.Unix()
+	return Ref(unix / 600)
+}
+
+type Shard struct {
+	sync.Mutex
+	ref Ref
+	m   map[SubKey]Stamp
```
replay: what does `m` stand for?

publish/kafka/keycache/shard.go
```diff
@@ -0,0 +1,104 @@
...
+	s.Lock()
+
+	// the amount to subtract of a stamp for it to be based on the new reference
+	subtract := newRef - s.ref
+
+	for subkey, stamp := range s.m {
+		// remove entry if it is too old, e.g. if:
+		// newRef - diff > "timestamp of the entry in 10minutely buckets"
+		// newRef - diff > ref + stamp
+		if uint32(newRef)-uint32(diff) > uint32(s.ref)+uint32(stamp) {
```
replay: this is evaluating `uint32(newRef)-uint32(diff)` on each iteration, but those values don't change and it could be done just once before the `for`

replay:
I think the locking of the key cache could be improved by 1) use RWMutex and take advantage of the many cases where only read lock is necessary and 2) probably could use `atomic` for the updates of the time stamps to not require write locks for that either

replay:
i think the locking could be improved a lot by using an RWMutex and also by using atomic.Store to get rid of a lot of write locks

publish/kafka/keycache/shard.go
```diff
@@ -0,0 +1,104 @@
...
+		// filling that into (2):
+		// newRef - diff <= ref + newRef - ref
+		// - diff <= 0
+		// diff >= 0
+		// 2) the result fits into a uint8. but since we decrease the amount, to a new >= 0 amount,
+		// we know it does
+
+		// we know subtract fits into a Stamp since we call Prune at least every 42 hours
+		s.m[subkey] = stamp - Stamp(subtract)
+		remaining++
```
replay: instead of maintaining a counter you could just get `len(s.m)` at the end of this loop

publish/kafka/keycache/keycache.go
woodsaj: I have done some benchmarks, lock contention is not going to be an issue.  But as @replay mentions a RWLock would probably be better.

Dieterbe: Re bad design / contention, I expect this highlevel, global lock to see little contention because the caches map is very small (O(num-orgs) ) and all operations on it are very short.
The underlying caches use sharded locks because they contain a lot more data, and the operations on it (specifically pruning) are more expensive and locking them for a long time can result in slowing down publishers, which I want to avoid.

publish/kafka/keycache/shard.go
Dieterbe: in principle, our code shouldn't have to copy, because no matter what happens to the key, after you've used it as an index in the map, it doesn't affect the map (see https://play.golang.org/p/dWtZ1086-Tx), I think because the map internally creates a copy.
However our problem is that SubKey is an array type, and i couldn't find a more elegant way to convert a key to a subkey, per https://stackoverflow.com/questions/30285680/go-lang-how-to-convert-slice-to-fixed-size-array

publish/kafka/keycache/shard.go
Dieterbe: map :p not very descriptive I know. I'll rename it to "data". slightly more descriptive

Dieterbe:
@replay 
> using atomic.Store to get rid of a lot of write locks
what exactly did you mean with this? there's Value.Store which I don't think is a good fit here, and a bunch of Store function for larger sized uints but not uint8

publish/kafka/keycache/shard.go
woodsaj: to be clear,  SubKey is a [15]byte and schema.Key is [16]byte.  There is no other way to convert a schema.Key to a SubKey except via a `copy()`.
Our benchmarking shows this keycache design (including this copy) is extremely memory efficient for both heap sizes and allocations.

replay:
@Dieterbe i was talking about https://github.com/raintank/tsdb-gw/pull/89/files#diff-b422505e2a3861be3f413444318d233bR48
Currently a lock on the whole shard is acquired to update a timestamp value, even if the subkey already exists. If instead of a `Mutex` this would be an `RWMutex`, you could check whether a sub key is already present with only a read lock (in the majority of cases it will probably be present) and then you could use `Value.Store()` to update the value, so as long as a subkey already exists it would not be necessary to acquire a write lock at all.

PR #90, opened 2018-04-16
only build binary once and use debian images

replay:
This is now reusing the binaries. But remember that this will most likely break `scripts/build_docker.sh` on work stations

### DIFF
--- a/cmd/cortex-gw/Dockerfile
+++ b/cmd/cortex-gw/Dockerfile
@@ -1,28 +1,12 @@
-ARG BASE_PATH=/go/src/github.com/raintank/tsdb-gw
-
-# build binaries inside an alpine3.7 container
-FROM golang:1.9.3-alpine3.7 AS build
-ARG BASE_PATH
-
-RUN apk --update add build-base linux-headers openssl-dev cyrus-sasl-dev bash python git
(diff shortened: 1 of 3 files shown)


cmd/cortex-gw/Dockerfile
```diff
@@ -1,28 +1,15 @@
...
-# as was used to build the binaries in
-FROM alpine:3.7
+FROM debian:stretch-slim
 LABEL author="Grafana Labs"
-ARG BASE_PATH
 
-RUN apk --update add ca-certificates
-RUN apk add openssl cyrus-sasl
+RUN apt-get update
+RUN apt-get install -y libssl1.1 libsasl2-2
```
woodsaj: libssl is only needed for librdkafka right?  so we dont need them in the cortex image

replay: right

jtlisi:
@replay `./scripts/build_docker.sh` works on my local. This looks good to me

replay:
@jtlisi sorry my comment was not precise. it does not break the execution of `scripts/build_docker.sh`, but it may result in that image not being able to start due to unsatisfied dynamic linker dependencies.

Dieterbe:
#101

PR #93, opened 2018-04-17
consolidate flags in main file for cortex-gw

jtlisi:


Dieterbe:
I think one can argue both ways about whether this is better or not.
For variables that are globals I think it makes sense to declare them in the file where they are actually used. If we were able to create multiple cortex readers and writers each with their own read and write url's this would make sense, but if we only have 1 anyway, we might as well stick to the simple approach and not have the needless indirections.
What's the argument for moving all the flags into one place? If it's seeing them in one place, then i would say that's what `cortex-gw --help` is for (or the config file)

Also, why do some of these flags say `cortex` in them? I think for all of them cortex is implied already.

PR #94, opened 2018-04-18
refactor flags in package file

jtlisi:
@Dieterbe I closed my last PR this one is more in line with what we are looking for given your suggestions.

### DIFF
--- a/publish/cortex/publish.go
+++ b/publish/cortex/publish.go
@@ -26,9 +26,9 @@ import (
 )
 
 var (
-	cortexWriteBPoolSize  = flag.Int("cortex-bpool-size", 100, "max number of byte buffers in the cortex write buffer pool")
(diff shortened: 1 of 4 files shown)


publish/cortex/publish.go
```diff
@@ -67,22 +67,21 @@ var (
 const maxErrMsgLen = 256
 
 // Init initializes the cortex reverse proxies
-func init() {
-	flag.Parse()
-
-	cortexURL, err := url.Parse(*cortexWriteURL)
+func Init() {
```
Dieterbe: why a non-standard init function?

Dieterbe:
this is now very similar to how we do it in MT. except in WT we call it ConfigProcess, what is here Init.
looks good to me, but don't forget to sync these changes to the config file...

jtlisi:
@Dieterbe Cool, config flags are updated so this should be all set

PR #97, opened 2018-04-20
validate compression; add lz4

Dieterbe:
https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md


replay:
is the `lz4` library getting added to the docker img? IIRC that file that you link to `CONFIGURATION` gets regenerated by `./configure`, if a lib is missing then it will remove that option from the available options. I assume the one that's posted in edenhill's repo has all potential options enabled

jtlisi:
No longer needed #101

PR #100, opened 2018-04-30
persister gw

jtlisi:
This PR adds a persister binary that can be used to persist metrics to the gateway. This allows metrics that are only sent once to be repeatedly sent to the service.

Persister Functional Path
Metric -> (tsdb-gw/cortex-gw)`<ingest endpoint>` -> persister ->  (tsdb-gw/cortex-gw)`/metrics`

This is designed in a way where it is up to the gateway to persist metrics. For example the datadog/intake enpoint.

@woodsaj I figured out the issue I was mentioning this morning, I wasn't setting the ID for the metrics before pushing them to the persister. After setting the ids on the persistor side it worked great


Dieterbe:
please update this PR to describe what this is about.

cmd/persister-gw/main.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/raintank/tsdb-gw/persister"
+	"github.com/raintank/tsdb-gw/util"
+	log "github.com/sirupsen/logrus"
+)
+
+/*
+Application: persister-gw
+
+persister-gw is stores and persists metrics to a configure publisher backend. It will store metrics
```
replay: i think the `is` seems unnecessary

ingest/datadog.go
```diff
@@ -89,6 +90,78 @@ func DataDogWrite(ctx *api.Context) {
...
+	var body []byte
+	var err error
+	if ctx.Req.Request.Header.Get("Content-Encoding") == "deflate" {
+		zr, err := zlib.NewReader(ctx.Req.Request.Body)
+		if err != nil {
+			log.Errorf("unable to decode json: %v", err)
+			ctx.JSON(500, err)
+			return
+		}
+		body, err = ioutil.ReadAll(zr)
```
replay: this line is kind of redundant with `:121`. the `body,err = ioutil.ReadAll(reader)` could be done after the `if/else`, so it would only need to be done once. furthermore, there seems to be no handler for the case there is an error, that error handler also only needs to be added once if you deduplicate the `ioutil.ReadAll()`

ingest/datadog.go
```diff
@@ -89,6 +90,78 @@ func DataDogWrite(ctx *api.Context) {
...
+	} else {
+		body, err = ioutil.ReadAll(ctx.Req.Request.Body)
+	}
+
+	var checks DataDogCheckPayload
+	err = json.Unmarshal(body, &checks)
+	if err != nil {
+		return
```
replay: would it be possible to write the error back to the sender? it might help the sender to fix their json

ingest/datadog.go
```diff
@@ -89,6 +90,78 @@ func DataDogWrite(ctx *api.Context) {
...
+type DataDogCheckPayload []struct {
+	Check     string   `json:"check"`
+	Host      string   `json:"host_name"`
+	Timestamp int64    `json:"timestamp"`
+	Status    float64  `json:"status"`
+	Message   string   `json:"message"`
+	Tags      []string `json:"tags"`
+}
+
+func DataDogCheck(ctx *api.Context) {
```
replay: the whole first ~25 lines of this function appear to be identical with the beginning of `DataDogSeries()`. Would it not be possible to deduplicated that into a function that takes `ctx` and returns either body or an `error`?

ingest/datadog.go
```diff
@@ -113,3 +186,80 @@ func createTagSet(host string, device string, ctags []string) []string {
...
+	} `json:"systemStats"`
+	Meta struct {
+		SocketHostname string `json:"socket-hostname"`
+		SocketFqdn     string `json:"socket-fqdn"`
+		Hostname       string `json:"hostname"`
+	} `json:"meta"`
+	Gohai string `json:"gohai"`
+}
+
+func (i *DataDogIntakePayload) GeneratePersistantMetrics(orgID int) []*schema.MetricData {
```
replay: afaik `Persistent` is with `e`

ingest/datadog.go
```diff
@@ -113,3 +186,80 @@ func createTagSet(host string, device string, ctags []string) []string {
...
+	if ctx.Req.Request.Header.Get("Content-Encoding") == "deflate" {
+		zr, err := zlib.NewReader(ctx.Req.Request.Body)
+		if err != nil {
+			log.Errorf("unable to decode json: %v", err)
+			ctx.JSON(500, err)
+			return
+		}
+		body, err = ioutil.ReadAll(zr)
+	} else {
+		body, err = ioutil.ReadAll(ctx.Req.Request.Body)
```
replay: this appears to be a copy-paste fo the above, so the same comments apply

metrics_client/metrics_client.go
```diff
@@ -0,0 +1,123 @@
...
+	data, err := msg.CreateMsg(mda, 0, msg.FormatMetricDataArrayMsgp)
+	if err != nil {
+		return 0, err
+	}
+
+	body := new(bytes.Buffer)
+	snappyBody := snappy.NewWriter(body)
+	snappyBody.Write(data)
+	snappyBody.Close()
+	req, err := http.NewRequest("POST", c.gwURL.String(), body)
```
replay: would it make sense to just store `gwURL` as a `string` type, so it wouldn't not need to be converted into a `string` on each push?

metrics_client/metrics_client.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	buf := make([]byte, 300)
+	n, _ := resp.Body.Read(buf)
+	log.Warningf("failed to persist data: http %d - %s", resp.StatusCode, buf[:n])
+
+	ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+
+	return resp.StatusCode, nil
```
replay: in this case we have no idea what value `resp.StatusCode` is, we only know that there was no error and it is not between 200 and 300. So, to not confuse users and maybe send invalid status codes, i guess it might be better to send some predefined error code here.

replay: what if the backend returned a valid code that was not between `200` and `300` and there was no error (f.e. `404`)?j

metrics_client/metrics_client.go
```diff
@@ -0,0 +1,123 @@
...
+		gwURL:    u,
+		gwAPIKey: cfg.APIKey,
+		client:   &http.Client{},
+	}, nil
+}
+
+func (c *Client) Push(metrics []*schema.MetricData) error {
+	now := time.Now()
+
+	statusCode, err := c.push(metrics)
```
replay: what if the backend returned a valid code that was not between `200` and `300` and there was no error (f.e. `404`)?

metrics_client/metrics_client.go
replay: what if the backend returned a valid code that was not between `200` and `300` and there was no error (f.e. `404`)? currently this would not be considered an error case

persister/persister.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		err = metricData.DecodeMetricData()
+		if err != nil {
+			log.Errorf("failed to unmarshal metricData. %s", err)
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		go p.Persist(metricData.Metrics)
```
replay: I'm surprised about that, because i thought usually in tsdb we'd only return a successful status once the data has successfully been written to the backend. or does that policy not apply here?

persister/persister.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (p *Persister) Persist(metrics []*schema.MetricData) {
+	log.Infof("persisting %v metrics", len(metrics))
+	p.Lock()
+	err := p.store.Store(metrics)
+	if err != nil {
+		log.Error(err)
+	}
+	p.metrics = append(p.metrics, metrics...)
```
replay: how are metrics getting cleared from `p.metrics`? they seem to always just get appended to it

persister/persister.go
replay: wouldn't it make sense to only return once the metrics have successfully been persisted, so we don't return `ok` when actually there was an error

persister/storage/gcp/storage.go
```diff
@@ -0,0 +1,139 @@
...
+// NewStorageClient returns a new StorageClient.
+func NewStorageClient(ctx context.Context, cfg Config) (storage.Storage, error) {
+	adminClient, err := bigtable.NewAdminClient(ctx, cfg.project, cfg.instance)
+	if err != nil {
+		return nil, err
+	}
+
+	tables, err := adminClient.Tables(context.Background())
+	if err != nil {
+		logrus.Fatalf("Could not fetch table list: %v", err)
```
replay: it seems kind of inconsistent that an error in `NewAdminClient()` results in an error being returned to the caller, while an error in `Tables()` results in `logrus.Fatalf`.

persister/persister.go
jtlisi: @replay I am looking into adding a delete endpoint. I was going to have it be a list of strings containing metrics id's but I realized that is not very useful because no one know the IDs of the metrics. So sending the metric name and tags might be the next best thing? When a delete is issued, my plan is to have the delete function return an updated slice of metrics from bigtable and  have it replace the old slice

persister/persister.go
```diff
@@ -0,0 +1,134 @@
...
+type Config struct {
+	orgID               int
+	interval            int
+	MetricsClientConfig metrics_client.Config
+	StorageClientConfig gcp.Config
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.IntVar(&cfg.orgID, "orgID", 1, "org id for the persister to manage")
```
woodsaj: as this service is primarily going to be used with Cortex, it needs to be multitenant.

persister/persister.go
woodsaj: i dont think a slice is going to work here.
When ever a dd-agent restarts it will re-send the host metadata metrics.  So each of those restarts will result in a new metric being added to this slice, which is not what we want.  An update from the dd-agent should replace the previous metrics from that agent.

so a map keyed on the orgId + host tag would probably be best.

ingest/datadog.go
```diff
@@ -113,3 +186,80 @@ func createTagSet(host string, device string, ctags []string) []string {
...
+	}
+
+	var info DataDogIntakePayload
+	err = json.Unmarshal(body, &info)
+	if err != nil {
+		return
+	}
+
+	if info.Gohai != "" {
```
woodsaj: why does info.Gohai need to be set for us to persist the metrics?

persister/persister.go
jtlisi: I updated the orgID to `bigtable-prefix` and left it blank. As a result the default mode is to retrieve metrics to persist from memory from all rows in the table. For persisting/deleting metrics in bigtable that functions regardless of whether the prefix is set. The prefix only determines which metrics will be regularly sent. If we are going to have this service also be multitenant it likely will require something more complex than loading the metrics into memory and sending the bulk payload on a timer. Any advice on how work is going to be distributed and executed? The rows are keyed on MetricData.ID so is there some partitioning code we can take advantage of?

jtlisi: Although I imagine that is something we can iterate on later.

woodsaj: Dont worry about shading yet.  We can add that later when we need it. A single persister will be able to handle 10's of 1000's of hosts before we need to shard the workload.

jtlisi:
@woodsaj I updated the intake struct to match all the fields of the intakes initial JSON. I will add all the fields that are sent multiple times as we iterate. Now the persister doesn't store the schema.MetricData it stores the json bytes of the intake payload. This should be good for now and we can iterate from here?

cmd/persister-gw/main.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/raintank/tsdb-gw/persister"
+	"github.com/raintank/tsdb-gw/util"
+	log "github.com/sirupsen/logrus"
+)
+
+/*
+Application: persister-gw
+
+persister-gw stores and persists metrics to a configure publisher backend. It will store metrics
```
woodsaj: > It will store metrics

either finish the sentence, or get rid of it.

ingest/datadog/intake.go
```diff
@@ -0,0 +1,238 @@
...
+		GoogleCloudPlatform []string `json:"google cloud platform,omitempty"`
+	}
+	Gohai string `json:"gohai"`
+	OrgID int    `json:"org-id"`
+}
+
+func (i *DataDogIntakePayload) GeneratePersistentMetrics() []*schema.MetricData {
+	metrics := []*schema.MetricData{}
+	tags := []string{
+		"agentVersion=" + serializeTag(i.AgentVersion),
```
woodsaj: i dont really like the idea of having lots of tags with the value "N/A".  if the value is not set, then we should just ommit the tag.
eg, 
```
tags := make([]string, 0)
sysInfoTags := map[string]string{
  "host": i.Meta.Hostname,
  "hostname": i.InternalHostname,
....
}
for k,v := range systemInfoTags {
  if v != "" {
    tags = append(tags, k+"="+v)
  }
}
```

ingest/datadog/intake.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		log.Errorf("unable to decode Gohai payload: %v", err)
+		return metrics
+	}
+
+	metrics = append(metrics,
+		&schema.MetricData{
+			Name: "system_platform_info",
+			Tags: []string{
+				"GOOARCH=" + serializeTag(g.Platform.GOOARCH),
```
woodsaj: as above, i dont really like the idea of having lots of tags with the value "N/A".

ingest/datadog/intake.go
```diff
@@ -0,0 +1,238 @@
...
+	info.OrgID = ctx.ID
+	err = json.Unmarshal(data, &info)
+
+	if err != nil {
+		ctx.JSON(400, fmt.Sprintf("unable to unmarshal request, reason: %v", err))
+		return
+	}
+
+	if info.Gohai != "" {
+		payload, err := json.Marshal(info)
```
woodsaj: You have missed the point of sending raw json to the persister.  The reason for sending raw json is because the DataDogIntakePayload struct does not capture all data sent.  
What you are doing here is to unmarshal the payload to the DataDogIntakePayload struct, then marshal it back to raw json, which will filter out all of the data that the DataDogIntakePayload doesnt explicitly describe.

so just send something like
```
type PersistPayload struct {
   OrgID int `json:"org-id"`   // used in rowKey
   Hostname string `json:"string"` // used in rowKey
   Raw []byte `json:"raw"`
}
```

```
if info.Gohai != "" {
  payload, err := json.Marshal(PersistPayload{OrgID: ctx.ID, Hostname: info.InternalHostname, RAW: data})
  ....
  err = persist.Persist(payload)
  ...
}
```

jtlisi:
@woodsaj I think this should do the trick. I made the requested changes and tested it locally

ingest/carbon/carbon.go
```diff
@@ -43,7 +43,7 @@ var (
 func init() {
 	flag.BoolVar(&Enabled, "carbon-enabled", false, "enable carbon input")
 	flag.StringVar(&addr, "carbon-addr", "0.0.0.0:2003", "listen address for carbon input")
-	flag.StringVar(&authPlugin, "carbon-auth-plugin", "file", "auth plugin to use. (grafana|file)")
+	flag.StringVar(&authPlugin, "carbon-auth-plugin", "grafana", "auth plugin to use. (grafana|file)")
```
woodsaj: this needs to be `file` by default.

scripts/build.sh
```diff
@@ -25,3 +25,9 @@ cd ../tsdb-usage
 go build -ldflags "-X main.GitHash=$VERSION" -o $BUILD_DIR/tsdb-usage
 cd ../cortex-gw
 go build -ldflags "-X main.GitHash=$VERSION" -o $BUILD_DIR/cortex-gw
+cd ../persister-gw
+go build -ldflags "-X main.GitHash=$VERSION" -o $BUILD_DIR/persister-gw
+
+# delete temporary build dir of librdkafka, since it is linked statically we
+# don't need it anymore
+echo rm -rf $TMP_DIR
```
Dieterbe: TMP_DIR is not set and we don't use librdkafka anymore.

scripts/config/persister-gw.ini
```diff
@@ -0,0 +1,10 @@
+addr="0.0.0.0:9001"
```
Dieterbe: this does not match what persister-gw stipulates in its main.

publish/publish.go
```diff
@@ -12,6 +13,9 @@ type Publisher interface {
 
 var (
 	publisher Publisher
+
+	// Persister allows pushing metrics to the Persistor Service
+	Persistor *metrics_client.Client
```
Dieterbe: spelling consistency? persistor vs persister

Issue #99, opened 2018-04-25
recent tsdb-gw with confluent becomes much slower under load


Dieterbe:
old = sarama
tsdb-gw = confluent (with GW_METRICS_BUFFER_MAX_MS: 50)

note how confluent is slower during normal operation, and apparently especially so when the cluster gets a bit more loaded.

![tsdbgw-latency-old-vs-new](https://user-images.githubusercontent.com/20774/39241978-d0c9f67c-4888-11e8-8cb8-b93b0ccf7ac5.png)

Dieterbe:
note that in our ops environment, publish latency mean went from 127 to 107ms  when we upgraded from old to new. (!) this was with the default of GW_METRICS_BUFFER_MAX_MS 100

Dieterbe:
branch reproduce-tsdbgw-sarama
```
./launch.sh docker-dev-custom-cfg-kafka
fakemetrics feed --gnet-addr http://localhost:9000/metrics --gnet-key not_very_secret_key --mpo 2000 --orgs 1
...
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock gaiaadm/pumba -- pumba --debug netem --tc-image gaiadocker/iproute2  --target $(docker inspect $(docker ps | grep raintank/kafka | awk '{print $1}') | sed -n 's#"IPAddress": "\(172.*\)",#\1#p') --duration 60s delay --time 10 --jitter 3 docker-dev-custom-cfg-kafka_tsdb-gw_1
```
https://snapshot.raintank.io/dashboard/snapshot/RgrInP9StXCwqy2fOwGYDySKur9onZqe?orgId=2

Dieterbe:
reproduce-tsdbgw-confluent 20kHz rate https://snapshot.raintank.io/dashboard/snapshot/aLkO1ZYOgOL3f8DKBYEn2FnCIKrJ2yxt?orgId=2

same, reproduce-tsdbgw-sarama
https://snapshot.raintank.io/dashboard/snapshot/hs0bi9z6pyHK5K4USBNlZhYEeLVGFmVz?orgId=2

Dieterbe:
now with snappy enabled

reproduce-tsdbgw-sarama: 2kHz, elevate latency, switch back latency, than 20kHz, and elevate latency again

https://snapshot.raintank.io/dashboard/snapshot/l3rr2uN2zSAD5oWxySrDh7GugJpWAi61?orgId=2

PR #101, opened 2018-05-02
switch back to sarama

replay:
This is replacing Confluent with Sarama again. i verified that it is functional, although i did not do much load testing yet. a lot of the code is copied from the version before the switch to confluent. 
It is now using Sarama version `1.16.0`, which is more recent than the one we have used before switching tsdb-gw to Confluent.


cmd/cortex-gw/Dockerfile
```diff
@@ -1,32 +1,14 @@
...
 COPY scripts/config/cortex-gw.ini /etc/gw/cortex-gw.ini
+COPY build/cortex-gw /usr/bin/cortex-gw
 COPY scripts/entrypoint.sh /usr/bin/
 
-# copy the built binaries from the build image
-COPY --from=build $BASE_PATH/build/cortex-gw /usr/bin/cortex-gw
-
 EXPOSE 80
 EXPOSE 443
-EXPOSE 8001
```
jtlisi: We still need to expose this port for prometheus metrics.

publish/kafka/publish.go
```diff
@@ -69,134 +56,91 @@ type Partitioner interface {
...
-	config.SetKey("max.in.flight", maxInFlight)
-	config.SetKey("queue.buffering.max.ms", bufferMaxMs)
-	config.SetKey("batch.num.messages", batchNumMessages)
-	config.SetKey("queue.buffering.max.messages", bufferMaxMsgs)
-
-	producer, err = kafka.NewProducer(&config)
+	partitioner, err = p.NewKafka(partitionScheme)
 	if err != nil {
-		log.Fatalf("failed to initialize kafka producer. %s", err)
+		log.Fatal(4, "failed to initialize partitioner. %s", err)
```
jtlisi: Since logrus is in use now, better to use log.Fatalf. Otherwise I don't think the variable interpolation will occur.

scripts/build.sh
```diff
@@ -1,23 +1,16 @@
...
-then
-	source scripts/build_deps.sh
-else
-	echo "not building librdkafka"
-fi
-
-# get git version
-GIT_VERSION=`git describe --always`
+# Disable CGO for builds.
+export CGO_ENABLED=0
```
jtlisi: We can probably remove this since CGO is disabled by default

tehlers320:
any more notes on why the switch back? We observed performance improvements in our Kafka cluster with the confluent lib.

tehlers320:
Found it https://github.com/raintank/tsdb-gw/issues/99

woodsaj:
other then @jtlisi's comments, this looks good to me. :+1:

replay:
@tehlers320 i replied to you on slack

Dieterbe:
`git rm -r vendor/github.com/edenhill`

publish/kafka/publish.go
replay: :+1: updated many of the log statements that didn't use it

Dieterbe:
do we still need all these libs?
```
$ git diff 0c201c80c3ce82d9807da3baba8495c4ac75bda8 HEAD
diff --git a/.circleci/config.yml b/.circleci/config.yml
index 39f4a73..2cb79d9 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,7 @@ jobs:
     docker:
       - image: circleci/golang:1.9.3
     steps:
+      - run: sudo apt-get install libssl-dev openssl libsasl2-2 libsasl2-dev zlib1g-dev
       - checkout
       - run: scripts/build.sh
       - persist_to_workspace:
@@ -17,9 +18,9 @@ jobs:
     docker:
       - image: circleci/golang:1.9.3
     steps:
+      - run: sudo apt-get install libssl-dev openssl libsasl2-2 libsasl2-dev zlib1g-dev
```

Dieterbe:
for completeness, the last sarama used pre-confluent was 1.12.0 (2017-05-08)

publish/kafka/publish.go
```diff
@@ -69,134 +56,91 @@ type Partitioner interface {
...
+	// Because we don't change the flush settings, sarama will try to produce messages
+	// as fast as possible to keep latency low.
+	config := sarama.NewConfig()
+	config.Producer.RequiredAcks = sarama.WaitForAll // Wait for all in-sync replicas to ack the message
+	config.Producer.Retry.Max = 10                   // Retry up to 10 times to produce the message
+	config.Producer.Compression = getCompression(codec)
+	config.Producer.Return.Successes = true
+	config.Producer.Flush.Frequency = flushFreq
+	config.Producer.Flush.MaxMessages = maxMessages
+	err = config.Validate()
```
Dieterbe: old code did this:
```
<       if err != nil {
<               log.Fatal(4, "failed to validate kafka config. %s", err)
<       }
```
should probably also validate here

publish/kafka/publish.go
```diff
@@ -69,134 +56,91 @@ type Partitioner interface {
...
-
 func init() {
 	flag.StringVar(&topic, "metrics-topic", "mdm", "topic for metrics")
 	flag.StringVar(&codec, "metrics-kafka-comp", "snappy", "compression: none|gzip|snappy")
 	flag.BoolVar(&enabled, "metrics-publish", false, "enable metric publishing")
 	flag.StringVar(&partitionScheme, "metrics-partition-scheme", "bySeries", "method used for paritioning metrics. (byOrg|bySeries)")
-	flag.IntVar(&maxInFlight, "metrics-max-in-flight", 1000000, "The maximum number of requests in flight per broker connection")
-	flag.IntVar(&bufferMaxMsgs, "metrics-buffer-max-msgs", 100000, "Maximum number of messages allowed on the producer queue. Publishing attempts will be rejected once this limit is reached.")
-	flag.IntVar(&bufferMaxMs, "metrics-buffer-max-ms", 100, "Delay in milliseconds to wait for messages in the producer queue to accumulate before constructing message batches (MessageSets) to transmit to brokers")
-	flag.IntVar(&batchNumMessages, "batch-num-messages", 10000, "Maximum number of messages batched in one MessageSet")
```
Dieterbe: so I don't know if you guys think i harp on this too much, but I think it's a useful best practice to always have the config files in sync with what the config params are that are really supported (and that we have matching default values). ./scripts/config/tsdb-gw.ini still uses these old values.

scripts/Dockerfile
```diff
@@ -1,31 +1,13 @@
...
-COPY . $BASE_PATH/
-RUN $BASE_PATH/scripts/build.sh
-
-
-# now build the production image based on the same alpine version
-# as was used to build the binaries in
-FROM alpine:3.7
-ARG BASE_PATH
-
-RUN apk --update add ca-certificates
```
Dieterbe: AFAIK we need the certificates in the final image. at least the last pre-confluent version had them.

publish/kafka/publish.go
```diff
@@ -252,71 +196,39 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
...
 		var buf []byte
 		for _, msg := range payload {
-			buf = msg.Value
-			if cap(buf) == 33 {
-				bufferPool33.Put(buf)
-			} else {
-				bufferPool.Put(buf)
-			}
+			buf, _ = msg.Value.Encode()
+			bufferPool.Put(buf)
```
Dieterbe: it's important to put the buffer in the most fit pool as we did before.
with this change, we will keep allocating anew whenever we want something from bufferPool33 (because this code never puts anything back in that pool)
and also cap-33 messages will be too small for the MetricData format, requiring another allocation.

scripts/build.sh
```diff
@@ -1,23 +1,13 @@
...
-if ! [ -d $PKG_CONFIG_PATH ] || [ -z $PKG_CONFIG_PATH ]
-then
-	source scripts/build_deps.sh
-else
-	echo "not building librdkafka"
-fi
-
-# get git version
-GIT_VERSION=`git describe --always`
+GITVERSION=`git describe --always`
```
Dieterbe: don't change this var name, it breaks the references to GIT_VERSION below

Dieterbe: or since in scripts/build_docker.sh we call it $VERSION let's do the same here

scripts/build.sh
```diff
@@ -1,23 +1,13 @@
 #!/bin/bash -e
 
-set -xe
+set -x
```
Dieterbe: why remove the `e` ? it seems useful to abort if any of the build commands fails.

publish/kafka/publish.go
```diff
@@ -252,71 +196,39 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
...
 		if err != nil {
 			return err
 		}
-
-		payload[i] = &kafka.Message{
-			TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: part},
-			Value:          data,
-			Key:            key,
+		payload[i] = &sarama.ProducerMessage{
+			Key:   sarama.ByteEncoder(key),
```
Dieterbe: as we discussed on slack. we no longer need to publish the keys. as we will build a new instance-upgrade mechanism by having tsdb-gw write to 2 topics when upgrading. for now we can't upgrade instances anyway because we had to derepcate the replicator tool in the MT repo.
still on board with this @woodsaj @replay ?
or if it's complicated to stop publishing the keys now (we just have to make sure partitions are being set correctly), we can also do this as a later step.

woodsaj: I think we can get rid of the keys, but not in this PR.  Lets do it after so we can more accurately test/benchmark the change.

Dieterbe: #103

Dieterbe:
looking good now, i'm just doing a bit of load and functionality testing

Dieterbe:
experiments done using the docker-dev-custom-cfg-kafka stack in MT repo
and `fakemetrics feed --gnet-addr http://localhost:9000/metrics --gnet-key not_very_secret_key --mpo 80000 --orgs 1`
I also confirmed with fake-metrics-data dashboard the data looks good.

# confluent
metrictank master, but updated to use gw 0.8.0-149-gcb39554 (latest gw with confluent, just before switch to sarama) and `metrics-buffer-max-ms = 50`
https://snapshot.raintank.io/dashboard/snapshot/y7PUwSGUZ3ObRLo21MjWFS66JFEaCWZ8?orgId=2
https://snapshot.raintank.io/dashboard/snapshot/GYcDDJGCZ7D8Pq4hvMezFvUHz0oiQtpb?orgId=2

# latest gw with sarama (code in this PR)
metrictank branch new-tsdbgw-sarama , gw config basically default (just with stats and publishing enabled basically)
https://snapshot.raintank.io/dashboard/snapshot/ghXNdeeVFkYaaOGStHX53zCfW5yqM2C7?orgId=2
https://snapshot.raintank.io/dashboard/snapshot/RRv39ulD1r6ciWDR1XR1jtw26PlEvAXo?orgId=2

PR #104, opened 2018-05-04
add standard and custom pprof endpoints

Dieterbe:
copied from metrictank
doesn't work yet
```
curl http://localhost:9000/debug/pprof/profile                                                                                                                         ⏎
404 page not found
```


jtlisi:
@woodsaj we could host the pprof endpoint on a seperate port similar to how to do with prometheus metrics. That way we could ideally standardize the pprof and metrics ports across all of our services

PR #105, opened 2018-05-04
tsdb-usage doesn't need storage-schemas.conf

Dieterbe:
because it generates all metrics by itself with known interval.

technically, tsdb-gw also only needs a storage-schemas.conf if we have input plugins enabled that don't enforce/know an interval (e.g. if only kafka input is enabled, we don't need it)
, but adding those conditionals to the code seems clumsy and fragile.


woodsaj:
Lgtm

PR #107, opened 2018-05-15
regularly force-refresh keycache to ensure periodic sending of MetricData

Dieterbe:
Metrictank users expect to be able to remove entries from the index,
and to have them re-appear if they keep sending data.
A simple way to achieve this is for tsdb-gw to regularly resend full
MetricData payloads, just in case users have deleted important index
entries.
This comes with 2 trade-offs:
* slighlty worse performance (due to the less efficient MetricData format)
* data may take an hour to re-appear


publish/kafka/keycache/keycache.go
```diff
@@ -9,36 +9,28 @@ import (
...
-	if pruneInterval.Minutes() < 10 {
-		panic("prune interval less than 10 minutes is useless due to resolution of internal bookkeeping")
-	}
+// each clearInterval, all shards will be wiped
+// (one at a time, spread out over the interval)
+func NewKeyCache(clearInterval time.Duration) *KeyCache {
 	k := &KeyCache{
-		pruneInterval: pruneInterval,
-		staleThresh:   Duration(int(staleThresh.Seconds()) / 600),
+		clearInterval: clearInterval / 256,
```
replay: what if `clearInterval` can not cleanly be divided by `256`? does `time.Duration` have a maximum granularity?
i don't think that's important because even if the value of `clearInterval` is not exactly applied due to rounding, it should not result in any issues, just curious.

Dieterbe: a time.Duration is a number of nanoseconds. the division may have a remainder but it'll be negligible (0-255 ns)

PR #108, opened 2018-05-19
expose counters as rates as well

Dieterbe:
because we like to aggregate across instances.
this makes that easier


DanCech:
Lgtm

PR #111, opened 2018-06-01
add require publisher authentication

jtlisi:
* add a middleware that checks whether a use has the publisher role
* user role is added to user struct
* `require-publisher` flag used to toggle feature

### DIFF
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -66,6 +66,16 @@ func RequireAdmin() macaron.Handler {
 	}
 }
 
+func (a *Api) RequirePublisher() macaron.Handler {
+	return func(ctx *models.Context) {
+		if !ctx.Role.IsPublisher() {
+			log.Errorf("user %v with role %v attempting to publish at %v", ctx.ID, ctx.Role, ctx.Req.RequestURI)
+			ctx.JSON(401, "Unauthorized to publish")
+			return
+		}
+	}
+}
+
 func (a *Api) Auth() macaron.Handler {
 	return func(ctx *models.Context) {
 		username, key, ok := ctx.Req.BasicAuth()
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 
+	"github.com/raintank/tsdb-gw/auth/gcom"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -11,11 +12,13 @@ var (
 	ErrInvalidCredentials = errors.New("invalid authentication credentials")
 	ErrInvalidOrgId       = errors.New("invalid orgId")
 	ErrInvalidInstanceID  = errors.New("invalid instanceID")
+	ErrInvalidRole        = errors.New("invalid authentication credentials, role unable to publish")
 
 	AdminKey  string
 	AdminUser = &User{
 		ID:      1,
 		IsAdmin: true,
+		Role:    gcom.ROLE_ADMIN,
 	}
 )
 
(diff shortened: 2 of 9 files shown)


DanCech:
The role will actually be called `MetricsPublisher` to avoid confusion, also api keys with that role should not be able to issue queries.

api/middleware.go
```diff
@@ -66,6 +66,46 @@ func RequireAdmin() macaron.Handler {
 	}
 }
 
+func RequirePublisher() macaron.Handler {
+	return func(ctx *models.Context) {
+		if !ctx.Role.IsPublisher() {
+			log.Errorf("user %v with role %v attempting to publish at %v", ctx.ID, ctx.Role, ctx.Req.RequestURI)
+			ctx.JSON(401, "Unauthorized to publish")
```
woodsaj: This should be a 403 error

api/middleware.go
```diff
@@ -66,6 +66,46 @@ func RequireAdmin() macaron.Handler {
...
+			return
+		}
+	}
+}
+
+func RequireViewer() macaron.Handler {
+	return func(ctx *models.Context) {
+		if !ctx.Role.IsViewer() {
+			log.Errorf("user %v with role %v attempting to view at %v", ctx.ID, ctx.Role, ctx.Req.RequestURI)
+			ctx.JSON(401, "Unauthorized to publish")
```
woodsaj: this should be a 403, and the message needs to be updated.

api/middleware.go
```diff
@@ -66,6 +66,46 @@ func RequireAdmin() macaron.Handler {
 	}
 }
 
+func RequirePublisher() macaron.Handler {
+	return func(ctx *models.Context) {
+		if !ctx.Role.IsPublisher() {
+			log.Errorf("user %v with role %v attempting to publish at %v", ctx.ID, ctx.Role, ctx.Req.RequestURI)
```
woodsaj: this should just be an info message.  It is not an application error if the users uses the wrong credentials

api/middleware.go
```diff
@@ -66,6 +66,46 @@ func RequireAdmin() macaron.Handler {
...
+			ctx.JSON(401, "Unauthorized to publish")
+			return
+		}
+	}
+}
+
+func RequireViewer() macaron.Handler {
+	return func(ctx *models.Context) {
+		if !ctx.Role.IsViewer() {
+			log.Errorf("user %v with role %v attempting to view at %v", ctx.ID, ctx.Role, ctx.Req.RequestURI)
```
woodsaj: this should just be an info message.  It is not an application error if the users uses the wrong credentials

jtlisi:
@DanCech @woodsaj this should be ready to merge if you have a chance to review

api/middleware.go
```diff
@@ -66,6 +66,46 @@ func RequireAdmin() macaron.Handler {
...
+	return func(ctx *models.Context) {
+		if !ctx.Role.IsViewer() {
+			log.Infof("user %v with role %v attempting to view at %v", ctx.ID, ctx.Role, ctx.Req.RequestURI)
+			ctx.JSON(403, "Unauthorized to view")
+			return
+		}
+	}
+}
+
+func (a *Api) GenerateHandlers(kind string, enforceRoles bool, datadog bool, handlers []macaron.Handler) []macaron.Handler {
```
woodsaj: handlers needs to be made a variadic paramater.
ie
```
func (a *Api) GenerateHandlers(kind string, enforceRoles bool, datadog bool, handlers ...macaron.Handler) []macaron.Handler {
```
then we can neaten up the code that calls this func.  Instead of calling
```
a.GenerateHandlers("read", enforceRoles, false, []macaron.Handler{a.PromStats("cortex-read"), cortex.Proxy})
```
you would call
```
a.GenerateHandlers("read", enforceRoles, false, a.PromStats("cortex-read"), cortex.Proxy)
```

PR #112, opened 2018-06-01
[WIP] filter by timerange

replay:
Not tested yet, going to test it. Also needed to update the macaron dependency to make the validator not blow up

### DIFF
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -1,8 +1,10 @@
 package api
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"path"
 	"strconv"
@@ -331,4 +333,14 @@ func Tracer(componentName string) macaron.Handler {
 		}
 		span.Finish()
 	}
+}
+
+func CaptureBody(c *models.Context) {
+	body, err := ioutil.ReadAll(c.Req.Request.Body)
+	if err != nil {
+		log.Error(3, "HTTP internal error: failed to read request body for proxying: %s", err)
+		c.PlainText(500, []byte("internal error: failed to read request body for proxying"))
+	}
+	c.Req.Request.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	c.Body = ioutil.NopCloser(bytes.NewBuffer(body))
 }
--- a/cmd/tsdb-gw/main.go
+++ b/cmd/tsdb-gw/main.go
@@ -11,8 +11,11 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/go-macaron/binding"
 	"github.com/grafana/globalconf"
+	"github.com/grafana/metrictank/api/middleware"
 	"github.com/grafana/metrictank/stats"
+	"github.com/raintank/dur"
 	"github.com/raintank/tsdb-gw/api"
 	"github.com/raintank/tsdb-gw/ingest"
 	"github.com/raintank/tsdb-gw/ingest/carbon"
@@ -53,6 +56,9 @@ var (
 	statsTimeout    = flag.Duration("stats-timeout", time.Second*10, "timeout after which a write is considered not successful")
 	tracingEnabled  = flag.Bool("tracing-enabled", false, "enable/disable distributed opentracing via jaeger")
 	tracingAddr     = flag.String("tracing-addr", "localhost:6831", "address of the jaeger agent to send data to")
+
+	// limitations
+	timerangeLimit = flag.String("timerange-limit", "", "define maximum timerange to serve queries for")
 )
 
 func main() {
@@ -109,7 +115,11 @@ func main() {
 		publish.Init(publisher)
 	}
 
-	if err := graphite.Init(*graphiteURL); err != nil {
+	var limit uint32
+	if len(*timerangeLimit) > 0 {
+		limit = dur.MustParseNDuration("timerange-limit", *timerangeLimit)
+	}
+	if err := graphite.Init(*graphiteURL, limit); err != nil {
 		log.Fatalf(err.Error())
 	}
 	if err := metrictank.Init(*metrictankURL); err != nil {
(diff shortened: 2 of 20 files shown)


cmd/tsdb-gw/validators.go
```diff
@@ -0,0 +1,38 @@
...
+
+	var to int64
+	if ft.to == 0 {
+		to = time.Now().Unix()
+	} else {
+		to = ft.to
+	}
+
+	if to-ft.from > timerangeLimitSec {
+		errs = append(errs, binding.Error{
```
DanCech: won't this fail in spectacular fashion if `from` and/or `to` are not unix timestamps?

replay: right, didn't think about that this also allows for other units. will fix that to make it accept the same formats as graphite has

Dieterbe: FWIW we can probably reuse the same library that MT uses to parse all graphite's formats

replay: @Dieterbe Good idea, unfortunately it is not exported (`getFromTo`). This exports it:

replay: nvm, that's pulling in tons of dependencies, and the function is short anyway, I'll just copy it over

replay:
Needed to a middleware-wrapper that duplicates the request body (the same way like the `proxyRetryTransport` does it, because binding a form reads the request body but in order to proxy a request the request body still needs to be readable.

api/middleware.go
```diff
@@ -66,6 +68,29 @@ func RequireAdmin() macaron.Handler {
...
+			body, err = ioutil.ReadAll(ctx.Req.Request.Body)
+			if err != nil {
+				ctx.JSON(500, err.Error())
+				return
+			}
+			ctx.Req.Request.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+			hasBody = true
+		}
+		ctx.Invoke(handler)
+		if hasBody {
```
Dieterbe: why do we do this again here?

also, did you see https://github.com/grafana/metrictank/blob/master/api/middleware/capturebody.go ? that approach looks a bit simpler

cmd/tsdb-gw/main.go
```diff
@@ -39,6 +40,9 @@ var (
 	graphiteURL   = flag.String("graphite-url", "http://localhost:8080", "graphite-api address")
 	metrictankURL = flag.String("metrictank-url", "http://localhost:6060", "metrictank address")
 
+	// limitations
+	timerangeLimit = flag.Duration("timerange-limit", 0, "define maximum timerange to serve queries for")
```
Dieterbe: whenever you add a cmdline flag, you should also update the config file to match. scripts/config/tsdb-gw.ini

also time.Duration has max unit `h` https://golang.org/pkg/time/#ParseDuration not very practical when dealing with large timeframes that are matters of days/weeks/months/years

that's why https://godoc.org/github.com/raintank/dur#ParseDuration exists

query/graphite/validators.go
```diff
@@ -0,0 +1,67 @@
...
+type FromTo struct {
+	From  string `json:"from" form:"from"`
+	Until string `json:"until" form:"until"`
+	To    string `json:"to" form:"to"` // graphite uses 'until' but we allow to alternatively cause it's shorter
+	Tz    string `json:"tz" form:"tz"`
+}
+
+func (ft FromTo) Validate(ctx *macaron.Context, errs binding.Errors) binding.Errors {
+	// we only care about queries to /graphite/render which specify a timerange
+	if len(ctx.Req.URL.Path) < 16 || ctx.Req.URL.Path[:16] != "/graphite/render" || ft.From == "" {
```
Dieterbe: strings.HasPrefix

api/middleware.go
replay: I did not see that, i'll just import it

cmd/tsdb-gw/main.go
replay: fixed

replay:
@Dieterbe fixed everything. Going to do some testing next

api/middleware.go
replay: The problem is that if we directly use that, and add a `.Body` property to the context to keep the body, then the graphite proxy (httputil.ReverseProxy) is still going to try to read `.Body` from the same property that has already been read. 
So I think a wrapper around the middle ware which reads the body is probably still the nicer solution, or not @Dieterbe ?

replay:
@Dieterbe did some testing now

This is without setting any value as time range limit:
```
mst@mst-nb1:~/documents/code/go/src/github.com/raintank/tsdb-gw$ curl -H 'Authorization: Bearer deadbeef' 'http://localhost:9000/graphite/render?target=tsdb-gw.stats.default.tsdb-gw.api.request.metrics_find.latency.max.gauge32&format=json&from=-2min' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     2  100     2    0     0     21      0 --:--:-- --:--:-- --:--:--    21
[]
```

Then I've set the time range limit to `2min`
```
mst@mst-nb1:~/documents/code/go/src/github.com/raintank/tsdb-gw$ curl -H 'Authorization: Bearer deadbeef' 'http://localhost:9000/graphite/render?target=tsdb-gw.stats.default.tsdb-gw.api.request.metrics_find.latency.max.gauge32&format=json&from=-2min' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     2  100     2    0     0     20      0 --:--:-- --:--:-- --:--:--    20
[]
mst@mst-nb1:~/documents/code/go/src/github.com/raintank/tsdb-gw$ curl -H 'Authorization: Bearer deadbeef' 'http://localhost:9000/graphite/render?target=tsdb-gw.stats.default.tsdb-gw.api.request.metrics_find.latency.max.gauge32&format=json&from=-3min' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    96  100    96    0     0  92307      0 --:--:-- --:--:-- --:--:-- 96000
[
  {
    "fieldNames": [
      "from",
      "to"
    ],
    "classification": "ValueError",
    "message": "Time range is too large"
  }
]

```

api/middleware.go
replay: Just importing didn't work, because MT and TSDB both use custom context structs. So I copied it. Also needed to slightly modify the graphite proxy to restore that captured body if it has been captured

PR #113, opened 2018-06-03
better logging and transport for cortex proxy

jtlisi:
This pipes std log output to logrus logger and uses a non default transport for the cortex reverse proxy.

### DIFF
--- a/publish/cortex/publish.go
+++ b/publish/cortex/publish.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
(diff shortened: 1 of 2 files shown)


Dieterbe:
> uses a non default transport for the cortex reverse proxy

what would be more interesting to know is : which setting is non-default and why do we need it?

jtlisi:
@Dieterbe 

I was looking into some 502 errors in cortex-gw. We use the httputil.ReverseProxy to forward requests to cortex. I found a google group post that has the same error https://groups.google.com/forum/#!topic/golang-nuts/oiBBZfUb2hM. Since the error is intermittent, the logging portion is more important. However, I would like to try this transport to see if this fixes it.

Dieterbe:
But we shouldn't just copy paste random code we find online and merge it, before understanding what it does.
When I compare this to the default transport I see a lot of differences being introduced all at once.

```
var DefaultTransport RoundTripper = &Transport{
        Proxy: ProxyFromEnvironment,
        DialContext: (&net.Dialer{
                Timeout:   30 * time.Second,
                KeepAlive: 30 * time.Second,
                DualStack: true,
        }).DialContext,
        MaxIdleConns:          100,
        IdleConnTimeout:       90 * time.Second,
        TLSHandshakeTimeout:   10 * time.Second,
        ExpectContinueTimeout: 1 * time.Second,
}
```

Have you considered them all?
for example: why use a non-Context aware dialer? why changing MaxIdleConns and KeepAlive all at once? why disable TLSHandshakeTimeout ?

jtlisi:
@Dieterbe I agree, I updated it to mostly default values, I didn't consider omitting the values. Only changing the Dialer initially to see if that has an effect.

Dieterbe:
@jtlisi when you close PR's please mention why. thanks

PR #115, opened 2018-08-21
Use the same buckets as underlying services

gouthamve:
Cortex uses different buckets as the gw which causes differences in the histogram numbers.


jtlisi:
@gouthamve sorry the dependencies on this repo are messy and it uses a legacy vendor tool. This looks great.

PR #116, opened 2018-08-30
add prometheus metrics endpoint and provide billing metrics

jtlisi:


cmd/tsdb-gw/main.go
```diff
@@ -189,3 +195,29 @@ func initRoutes(a *api.Api, enforceRoles bool) {
 	a.Router.Any("/prometheus/write", a.GenerateHandlers("write", enforceRoles, false, ingest.PrometheusMTWrite)...)
 	a.Router.Post("/metrics/delete", a.GenerateHandlers("write", enforceRoles, false, metrictank.MetrictankProxy("/metrics/delete"))...)
 }
+
+type metricsServer struct {
```
gouthamve: Hmm, this exists in `cortex-gw` also, can we put this in util?

publish/publish.go
```diff
@@ -28,6 +38,10 @@ func Init(p Publisher) {
 }
 
 func Publish(metrics []*schema.MetricData) error {
+	if len(metrics) < 1 {
```
gouthamve: Hmm, `== 0` is a more common pattern I think.

publish/publish.go
```diff
@@ -1,11 +1,21 @@
...
 	"github.com/raintank/tsdb-gw/metrics_client"
 	log "github.com/sirupsen/logrus"
 	schema "gopkg.in/raintank/schema.v1"
 )
 
+var (
+	ingestedMetrics = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "gateway",
+		Name:      "samples_ingested_total",
+		Help:      "Number of samples datapoints ingested",
```
Dieterbe: what's a samples datapoints ?

Dieterbe:
don't forget to update ./scripts/config/tsdb-gw.ini

jtlisi:
@gouthamve Each metrictank tenant has their own gateway so there is not need for a tenant id.

cmd/tsdb-gw/main.go
jtlisi: Will do

publish/publish.go
jtlisi: :+1:

publish/publish.go
jtlisi: Each sample is a <value,timestamp>. We are just going to rate and sum across all the gateways for a org to get their ingestion rate.

scripts/config/tsdb-gw.ini
```diff
@@ -55,3 +55,6 @@ tracing-addr = localhost:6831
 
 # limitations
 timerange-limit =
+
+# prometheus instrumentation
+metrics-addr = localhost:8001
```
Dieterbe: should match the default value

jtlisi: :+1:

PR #118, opened 2018-09-18
Use dep for vendoring, add opentracing instrumentation.

tomwilkie:


scripts/vendor_health.sh
```diff
@@ -1,24 +1,14 @@
...
+set -eu
 
 if ! grep -q $GOPATH/bin <<< $PATH ; then
   export PATH="$PATH:$GOPATH/bin"
 fi
 
-if ! which govendor >/dev/null; then
-	go get github.com/kardianos/govendor || exit 1
+if ! which dep >/dev/null; then
+	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
Dieterbe: MT runs `go get -u github.com/golang/dep/cmd/dep` which works fine.
https://github.com/grafana/metrictank/blob/master/scripts/qa/vendor.sh

tomwilkie:
This doesn't work, and the macaron router is so convoluted that I can't figure out why.

scripts/vendor_health.sh
tomwilkie: From https://github.com/golang/dep:
> It is strongly recommended that you use a released version.

Dieterbe:
I think it's important for a PR that changes vendored dependencies to be more clear about those changes. I created a new PR, see above link. I also didn't have any issues with macaron.

PR #124, opened 2018-10-22
carbon input: upon any error: don't just return, also close conn

Dieterbe:
err could be timeout or something else


Dieterbe:
note: this code will soon be replaced with the plugin from carbon-relay-ng

replay:
I think this could possibly be the bug that affects a customer every once in a while. If those connections don't get closed and just linger, could the process reach some resource limit which doesn't allow it to create any additional connections anymore?
IIRC the problem that we've seen with one specific customer was that after multiple weeks/months of running tsdb-gw it suddenly just stops accepting additional connections, without any error. a restart fixes this.

Dieterbe:
> it suddenly just stops accepting additional connections

how sure are we of this?
i think perhaps what the bug could be here is that we stop using the conn upon a read error, but perhaps the clients keep using it and not creating a new one.

ingest/carbon/carbon.go
```diff
@@ -171,9 +171,10 @@ func (c *Carbon) handleConn(conn net.Conn, shutdown chan struct{}, wg *sync.Wait
 		line, err := reader.ReadBytes('\n')
 
 		if err != nil {
-			if io.EOF != err {
+			if err != io.EOF {
 				log.Errorln(err.Error())
 			}
+			conn.Close()
```
woodsaj: Should we just move this up to the `defer` on line #152

replay:
@Dieterbe answered on slack

Dieterbe:
deprecated in favor of #127

PR #125, opened 2018-10-23
Sarama kafka version

woodsaj:
from  https://github.com/Shopify/sarama/blob/v1.19.0/config.go#L324-L330

```
The version of Kafka that Sarama will assume it is running against.
Defaults to the oldest supported stable version. Since Kafka provides
backwards-compatibility, setting it to a version older than you have
will not break anything, although it may prevent you from using the
latest features. Setting it to a version greater than you are actually
running may lead to random breakage.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: /go/src/github.com/raintank/tsdb-gw
     docker:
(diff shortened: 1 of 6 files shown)


Dieterbe:
scripts/config/tsdb-gw.ini must be updated accordingly

PR #126, opened 2018-10-23
Customise the number of connections to each host

gouthamve:
Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

### DIFF
--- a/publish/cortex/publish.go
+++ b/publish/cortex/publish.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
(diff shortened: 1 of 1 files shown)


Dieterbe:
> Still need to test it out.

i'll wait for further news then before merging ?

gouthamve:
screen shot 2018-10-23 at 12 44 25

The graph for the last 6 hours. The latency dropped quite a lot after the latest deploy. While the general 5xx have reduced (nothing in the last `1h`), we do have sporadic spikes with a new error:
```
2018/10/23 16:42:58 http: proxy error: EOF
2018/10/23 16:42:58 http: proxy error: EOF
```

gouthamve:
Hmm, irrespective of the new error, I think this should be merged. It's much more stable now. And the latency is muuuuch better. I'll open another PR for the error detected.

gouthamve:
This is causing super bad load-balancing for us though.

See this blog by Tom: https://www.weave.works/blog/turtles-way-http-grpc/

> We disabled persistent connections in AuthFE to downstream services as the combination of Go’s HTTP client and kubeproxy meant we weren’t getting good load balancing. This was quite critical for Scope’s query performance, as Scope queries are implausibly CPU intensive.

PR #131, opened 2018-10-26
refactor token and instance caches. fixes #130

### LINKED ISSUE #130: reduce latency due to authenticating against grafana.com
every request is to tsdb-gw/cortex-gw needs to be authenticated.   Authentication requires validating tokens against grafana.com, and for instance auth, then validating that the customer has an instance with the provided instanceId (provided as the username).

So we dont overload grafana.com we cache lookups.

However there are 2 main issues with the current design:

1) the default cache time is only 5minutes.  We should make it at least 1h
2) when a cached key hits its TTL, we re-validate the key against grafana.com before returning.  This can add a few hundred ms of latency to the request. More for 'grafana-instance' auth (cortex-gw) as the instance info will expire at the same time as the key and also need re-validation.

To address these i propose that:

If a key/instance is cached, we always consider it to be valid and immediately return.
Use a background goroutine to periodically re-validate keys.  If the key becomes invalid, then we mark it as such (the caches already support caching an invalid key).
We will also need to keep a record of when the key was last seen in a request so we can remove entries from the cache that are no longer being seen.

woodsaj:
Updating entries in the cache has been made async. This ensures that
1) requests are not blocked waiting for re-validation of tokens and
   instances
2) we dont get into situations where we flood grafana.com with
   validation requests

Instead of updating the cached entries when they are past their TTL,
We run a background tasks that scans over the cached items looking
for expired entries and re-validating them.

On reads of the cache, we update a lastRead timer.  If this timer
falls behind the last validation then we know that the cached item
is no longer being read, so we purge it from the cache.

To handle the background tasks, the AuthPlugin interface has been updated
to ensure plugins have a "Stop()" method.

Finally the auth-valid-ttl and auth-invalid-ttl cmdline args have been
replaced with a single 'auth-cache-ttl' arg, which defaults to 1hour.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -606,6 +606,14 @@
   revision = "6881fee410a5daf86371371f9ad451b95e168b71"
 
 [[projects]]
+  branch = "master"
+  digest = "1:5325f65188f8a9559c8f20d703b2b2684333a5e03515b6c1fbf09ffc710415af"
+  name = "golang.org/x/sync"
+  packages = ["singleflight"]
+  pruneopts = "NUT"
+  revision = "1d60e4601c6fd243af51cc01ddf169918a5407ca"
+
+[[projects]]
   digest = "1:5676bcdf4687d6daa95eefa77eb053ed722934677159274f40b75193f3e01cb2"
   name = "golang.org/x/sys"
   packages = ["unix"]
@@ -774,6 +782,7 @@
     "github.com/uber/jaeger-client-go/config",
     "github.com/uber/jaeger-client-go/log",
     "golang.org/x/net/context/ctxhttp",
+    "golang.org/x/sync/singleflight",
     "gopkg.in/ini.v1",
     "gopkg.in/macaron.v1",
   ]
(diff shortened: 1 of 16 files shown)


Dieterbe:
Is there a risk of flood when reqs come in for a previously unseen key ? There's packages designed to address that (groupcache which uses singleflight )

DanCech:
Looks good, I have the same concern @Dieterbe does

woodsaj:
Good call @Dieterbe.  i have wrapped the ValidateInstance/ValidateToken calls using [golang.org/x/sync/singleflight](https://godoc.org/golang.org/x/sync/singleflight)

woodsaj:
> There is still a small race between the singleflight block in ValidateToken and the cache being updated, but things are far better than they were.

Yes, but due to the latency of sending a validation to grafana.com, the worst case scenario is that we send 2 requests instead of the desired 1.   still a huge improvement to the on the 100's of requests to g.com we are sending today.

We could address this by adding to the cache within the singleflight.Do call, but i just want to get this out and reap the benefits.

PR #136, opened 2018-11-07
enforce auth-valid-instance-type when flag is set

jtlisi:
Signed-off-by: Jacob Lisi <jacob.t.lisi@gmail.com>


auth/gcom/roles.go
```diff
@@ -47,8 +48,17 @@ type SignedInUser struct {
...
+	InstanceType string `json:"type"`
 }
+
+type InstanceType string
+
+const (
+	Graphite       InstanceType = "graphite"
+	GraphiteShared InstanceType = "graphite-shared"
+	Cortex         InstanceType = "cortex"
+)
```
DanCech: Are these used anywhere?

auth/gcom/auth.go
```diff
@@ -211,6 +212,11 @@ func validateInstance(instanceID, token string) error {
 		return ErrInvalidInstanceID
 	}
 
+	if validInstanceType != "" && validInstanceType != instance.InstanceType {
+		log.Errorf("Auth: instanceType returned from grafana.com doesnt match required instanceType. %s != %s", instance.InstanceType, validInstanceType)
```
woodsaj: This shouldnt be logged as an error.  Probably should be debug, maybe info

PR #139, opened 2018-12-05
support auth-valid-instance-type=logs

DanCech:


DanCech:
Depends on https://github.com/raintank/grafana-net/pull/516

auth/gcom/auth.go
```diff
@@ -176,7 +176,15 @@ func ValidateInstance(cacheKey string) error {
 }
 
 func validateInstance(instanceID, token string) error {
-	req, err := http.NewRequest("GET", fmt.Sprintf("%s/api/hosted-metrics/%s", authEndpoint, instanceID), nil)
+	var instanceUrl string
+
+	if validInstanceType == "logs" {
```
woodsaj: would this be cleaner with something like
```
resource := "hosted-metrics"
if validInstanceType == "logs" {
  resource = "hosted-logs"
}
instanceUrl = fmt.Sprintf("%s/api/%s/%s", authEndpoint, resource, instanceID)
```

DanCech: I thought about that and figured that it was easier to follow what was going on with them separated, happy to change if anyone has strong feelings about it.

tomwilkie:
I don't know this code too well - @jtlisi would be best to review.  But LGTM!

auth/gcom/auth.go
```diff
@@ -191,7 +199,7 @@ func validateInstance(instanceID, token string) error {
 	body, err := ioutil.ReadAll(res.Body)
 	res.Body.Close()
 
-	log.Debugf("Auth: hosted-metrics response was: %s", body)
+	log.Debugf("Auth: %s response was: %s", authEndpoint, body)
```
woodsaj: I think this should print the full "instanceUrl" instead of just the authEndpoint.

PR #140, opened 2018-12-14
track samples_ingested_total per org

woodsaj:


publish/publish.go
```diff
@@ -1,19 +1,21 @@
...
 )
 
 var (
-	ingestedMetrics = promauto.NewCounter(prometheus.CounterOpts{
+	ingestedMetrics = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "gateway",
 		Name:      "samples_ingested_total",
 		Help:      "Number of samples ingested",
-	})
+	}, []string{"org"})
```
jtlisi: Can this tag be renamed to orgID?

woodsaj: depending on the auth plugin being used, this could be an OrgId or an InstanceId.   "org" is what cortex uses and i think we should be consistent.

Issue #121, opened 2018-10-12
report back invalid metrics to clients


Dieterbe:
there's a larger issue which needs documenting, which is:
1) tsdbgw rejects requests to /metrics if they contain any invalid metrics (even if N-1 are valid)
2) carbon-relay-ng will keep retrying requests that didn't succeed.

the combination of the above means that any invalid data blocks crng and halts your pipeline.

this led to https://github.com/graphite-ng/carbon-relay-ng/pull/292 but it is not enough:

A) we may always run into situations where a newer tsdbgw is more strict than an older carbon-relay-ng (which we can't force to be upgraded as they run on customer's premises), so we can't prevent the failure mode
B) customers/clients need better insights into which metrics were invalid.

so we're thinking of these 3 phases:

1) upon invalid data, reply http 202 and mention which series were invalid. this means any carbon-relay-ng, in particular older versions will continue streaming instead of choking. though users won't be able to see which metrics were invalid yet. but our HM dashboard should show the metric showing invalid data.
the response should be extensible (e.g. json with an entry for invalid data)

2) tsdb-gw should expose something similar to badMetrics in carbon-relay-ng, so we can see which metrics were invalid (either internally or in another service)

3) ability for customers to see reports of bad data based on 2

at some point we should also update carbon-relay-ng to log warnings for all invalid data.
probably between step 1 and 2.

woodsaj:
we should just return the bad metrics in the response from tsdb-gw to carbon-relay-ng.  Carbon-relay-ng can then log them.  Also, the response code should be 200.  202 is for asynchronous write requests, which is not what we do.   We only respond when kafka acknowledges that metrics have been ingested.
Even if metrics are invalid i think a 200 status code is still correct. Because the request is being handled successfully by tsdb-gw.

But i think we need to also return a response body and that is where we should report errors.

eg. if all metrics are valid we would return
```
{
    "errors": false,
    "received": 1000,
    "discarded": 0,
}
```
if there are errors we would return something like
```
{
    "errors": true,
    "received": 1000,
    "discarded": 2,
    "errors": [
        {
            metric: <MetricData>,
            reason: "interval must be > 0"
        },
        {
            metric: <MetricData>,
            reason: "invalid tag format"
        }
    ]
}
```
If there is a high number of errors we should only return the first N to keep the payload size reasonable.

Dieterbe:
[...]

PR #144, opened 2019-01-08
issue 121: don't reject ingestion requests upon invalid data

Dieterbe:
see #121 
see https://github.com/graphite-ng/carbon-relay-ng/pull/338 for corresponding crng PR

### DIFF
--- a/ingest/metrics.go
+++ b/ingest/metrics.go
@@ -116,6 +116,14 @@ func metricsJson(ctx *models.Context) {
 
 	toPublish := make([]*schema.MetricData, 0, len(metrics))
 	toPublish, resp := prepareIngest(ctx, metrics, toPublish)
+
+	select {
+	case <-ctx.Req.Context().Done():
+		ctx.Error(499, "request canceled")
+		return
+	default:
+	}
+
 	err = publish.Publish(toPublish)
 	if err != nil {
 		log.Errorf("failed to publish metrics. %s", err)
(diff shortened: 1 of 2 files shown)


ingest/metrics.go
```diff
@@ -34,128 +45,179 @@ func Metrics(ctx *models.Context) {
...
-		if err != nil {
-			log.Errorf("unable to read request body. %s", err)
-		}
-		metrics := make([]*schema.MetricData, 0)
-		err = json.Unmarshal(body, &metrics)
-		if err != nil {
-			ctx.JSON(400, fmt.Sprintf("unable to parse request body. %s", err))
-			return
-		}
+	body, err := ioutil.ReadAll(ctx.Req.Request.Body)
```
woodsaj: given ctx.Req.Request.Body is an io.Reader, should we just use json.NewDecoder.

with the approach documented at https://golang.org/pkg/encoding/json/#example_Decoder_Decode_stream we could also decode the MetricData objects as a stream and avoid having to load the entire batch into memory first.

Dieterbe: you're right. but if we want a centralized metrics processing and a json stream decoder, we also need to do something similar for messagepack, which means working on the schema package, refactoring more, etc.
I think better for now let's leave it as is, we have lower hanging fruit.

Dieterbe:
@woodsaj PTAL . i added some more commits. the one most interesting I think is 14f1ea4

PR #146, opened 2019-01-10
Fix race by removing global variable

gouthamve:
Fixes: https://github.com/raintank/tsdb-gw/issues/142


gouthamve:
The biggest issue is that there are a lot of goroutines reading `cacheTTL` spawned in the tests while the tests itself try to modify the variable, for example, here: https://github.com/raintank/tsdb-gw/blob/master/auth/gcom/auth_test.go#L335

PR #150, opened 2019-01-17
auth: Add ability to validate cluster name

gouthamve:
Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

### DIFF
--- a/auth/gcom/auth.go
+++ b/auth/gcom/auth.go
@@ -23,6 +23,7 @@ func init() {
 	flag.DurationVar(&defaultCacheTTL, "auth-cache-ttl", defaultCacheTTL, "how long auth responses should be cached")
 	flag.Var(&validOrgIds, "auth-valid-org-id", "restrict authentication to the listed orgId (comma separated list)")
 	flag.StringVar(&validInstanceType, "auth-valid-instance-type", "", "if set, instance validation while fail if the type attribute of an instance does not match. (graphite|graphite-shared|prometheus|logs)")
+	flag.StringVar(&validClusterName, "auth-valid-cluster-name", "", "if set, instance validation while fail if the cluster attribute of an instance does not match.")
 }
 
 type int64SliceFlag []int64
@@ -51,6 +52,7 @@ var (
 	authEndpoint      = "https://grafana.com"
 	validOrgIds       = int64SliceFlag{}
 	validInstanceType string
+	validClusterName  string
 
 	// global HTTP client.  By sharing the client we can take
 	// advantage of keepalives and re-use connections instead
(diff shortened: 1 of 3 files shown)


gouthamve:
This adds the ability to limit the valid instances to the cluster the gateway is deployed in.

auth/gcom/auth.go
```diff
@@ -23,6 +23,7 @@ func init() {
 	flag.DurationVar(&defaultCacheTTL, "auth-cache-ttl", defaultCacheTTL, "how long auth responses should be cached")
 	flag.Var(&validOrgIds, "auth-valid-org-id", "restrict authentication to the listed orgId (comma separated list)")
 	flag.StringVar(&validInstanceType, "auth-valid-instance-type", "", "if set, instance validation while fail if the type attribute of an instance does not match. (graphite|graphite-shared|prometheus|logs)")
+	flag.StringVar(&validClusterName, "auth-valid-cluster-name", "", "if set, instance validation while fail if the cluster attribute of an instance does not match.")
```
Dieterbe: IMHO always keep config files in sync with built in options. including their description and default value

auth/gcom/auth.go
```diff
@@ -225,6 +227,11 @@ func validateInstance(instanceID, token string) error {
 		return ErrInvalidInstanceType
 	}
 
+	if validClusterName != "" && validClusterName != instance.ClusterName {
```
jtlisi: @gouthamve could we update this to always log if the cluster name is different, or add a dry run feature. I'm worried we might end up dropping data because something is misconfigured.

DanCech: This should be using the `clusterSlug` or `clusterId` attribute, `clusterName` is the presentation name and shouldn't be relied on not to change.

auth/gcom/auth.go
```diff
@@ -221,10 +230,17 @@ func validateInstance(instanceID, token string) error {
...
 
 	if validInstanceType != "" && validInstanceType != instance.InstanceType {
+		validationFailed.WithLabelValues("instance").Inc()
 		log.Infof("Auth: instanceType returned from grafana.com doesnt match required instanceType. %s != %s", instance.InstanceType, validInstanceType)
 		return ErrInvalidInstanceType
 	}
 
+	if validClusterID != 0 && validClusterID != instance.ClusterID {
+		validationFailed.WithLabelValues("cluster").Inc()
+		log.Infof("Auth: clusterName returned from grafana.com doesnt match required clusterID. %d != %d", instance.ClusterID, validClusterID)
```
DanCech:
```suggestion
		log.Infof("Auth: cluster id returned from grafana.com doesnt match required clusterID. %d != %d", instance.ClusterID, validClusterID)
```

gouthamve:
I added a dry run feature. But the issue here is that it's only logged once. After it has been validated, it's cached and hence not logged for every request. This is annoying as it'll quickly get lost in the huge volume of the request logs.

PR #155, opened 2019-04-03
Add support for websockets

sandlis:
To support websockets with reverse proxy, response writer needs to satisfy `http.Hijacker` interface.
Added this method to `TracingResponseWriter` which in turn calls underlying Hijack method implemented by `macaron.ResponseWriter`

This is needed to support live tailing of logs from loki

### DIFF
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -1,10 +1,12 @@
 package api
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"path"
 	"strconv"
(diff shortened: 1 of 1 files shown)


Dieterbe:
lgtm though i'm not very familiar with this, so i recommend another person to look as well

sandlis:
> lgtm though i'm not very familiar with this, so i recommend another person to look as well

Thanks @Dieterbe for having a look!
Tom has also tagged Jacob to have a look at this.

sandlis:
@jtlisi Yes, I have tested it locally. Works fine.

tomwilkie:
A unit test for this would be good, we'll break it in the future if we're not careful.

PR #158, opened 2019-04-10
Unit tests for websockets

sandlis:
Adding tests for websockets which would help make sure any newly added middlewares pass these test

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -8,7 +8,6 @@ import (
 	"net/http/httptest"
 	"net/http/httputil"
 	"net/url"
-	"strconv"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


api/api_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"net/http/httptest"
+	"net/http/httputil"
+	"net/url"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/raintank/tsdb-gw/api/models"
+	"github.com/raintank/tsdb-gw/auth"
+	"golang.org/x/net/http/httpguts"
```
Dieterbe: interesting package..

api/api_test.go
```diff
@@ -0,0 +1,147 @@
...
+		if !bs.Scan() {
+			t.Errorf("backend failed to read line from client: %v", bs.Err())
+			return
+		}
+		fmt.Fprintf(c, "backend got %q\n", bs.Text())
+	}))
+
+}
+
+func TestReverseProxyWebSocketWithHijacker(t *testing.T) {
```
Dieterbe: as a reader, this is hard to follow. some comments would be helpful.
what are we actually testing for here? why are we adding routes here and not testing existing ones? what is the X-Scope-OrgID header?

Dieterbe:
looks pretty good, just a bit hard to follow for me. see comments.

api/api_test.go
sandlis: @Dieterbe Added comments and removed X-Scope-OrgID since its not required.
Please check whether it looks good.
I am adding new route since different commands have different routes, wanted to make it a generic test.

PR #161, opened 2019-04-12
Upgrade go to 1.12 and some packages used for testing

sandlis:
Support for websockets using reverse proxy was introduced in go 1.12
Using go 1.12 image in circle ci and upgrading some packages which were broken when using go 1.12


woodsaj:
@sandlis can you force push to trigger CI tasks to run.  I have updated the circleci config so that PRs from forks trigger builds

sandlis:
@woodsaj `build_accept_deploy` passed, others are still blocked.

woodsaj:
i fixed the repo settings.  after adding the circleci "checks" integration, all checks are listed with the single "build_accept_deploy" name.

PR #163, opened 2019-05-14
slightly better readme

Dieterbe:
@jtlisi can you fill in the persister-gw part and @woodsaj can you do the auth part

intention is not to have very rich documentation - cause it's bound to get out of date anyway, but at least document the most important things


Dieterbe:
woodsaj mentioned he won't have time for this in a while.
i want to give @jtlisi another day or two. if no response, will merge this as is

PR #164, opened 2019-05-30
add endpoint for mt bulk importer

replay:
this adds an optional endpoint to tsdb-gw which can be used to bulk import data into metrictank's backend store


Dieterbe:
if auth works (c.ID set as expected) then LGTM

PR #165, opened 2019-06-04
measure min/max timestamp seen for carbon/http ingest plugins. fix #162

### LINKED ISSUE #162: monitor data timestamps "sync" with wall clock
I would like to have a metric showing how far behind or ahead incoming metrics are wrt wall clock.
why?
1) it's useful for debugging/general understanding
2) it helps knowing how much overlap we need between old and new write pods during migrations (see #160)

instead of reporting, we may as well report the min and max timestamps seen at each point.
that provides more information, and it should be easy to derive the min/max lag numbers from that.

Dieterbe:


fkaleo:
LGTM

PR #166, opened 2019-06-10
create a separate route handler for Loki tail endpoint

slim-bean:
This should help clear up the issue we are seeing with query stats for loki, as all the websocket connections are being tracked through the same histogram as query requests.


slim-bean:
closing this because this project is being replaced with grafanacloud-tools

PR #167, opened 2019-06-10
remove editor project files

slim-bean:
we should keep editor project files out of the source code, I would suggest setting up a global gitignore for the `.idea/` folder, that was the only way I could keep myself form accidentally checking in these files :)


slim-bean:
closing this because this project is being replaced with grafanacloud-tools

PR #171, opened 2019-07-02
Timestamp stat per org

Dieterbe:


woodsaj:
I really dont like the idea of forking the metrictank stats library.  Metrictank is also going to need the stats library updated to support tags.  Implementing that support should be done once, in one place.

Dieterbe:
sure. we'll merge them back once we validate the approach

ingest/metrics.go
```diff
@@ -79,6 +78,8 @@ func prepareIngest(ctx *models.Context, in []*schema.MetricData, toPublish []*sc
 		if !ctx.IsAdmin {
 			m.SetId()
 		}
+		suffix := fmt.Sprintf(";org=%d", m.OrgId)
+		metricsTimestamp := stats.NewRange32("metrics.timestamp.http", suffix) // min/max timestamps seen in each interval
```
woodsaj: calling `stats.NewRange32()` on every datapoint is extremely inefficient. `NewRange32()` calls `registry.getOrAdd` which uses reflect.

We should just keep a map of metrics, key'ed by the orgId. eg`map[int]*stats.Range32`

Also, we should optimise for `ctx.Admin` not being true, which will allow us to get the per Org stat once based on the ctx.ID, and not have to look it up for each datapoint in the batch received.

woodsaj:
> sure. we'll merge them back once we validate the approach

Getting the right approach is going to take lots of time and planning.  I think it is best that we just dont use tags right now.

Dieterbe:
> Getting the right approach is going to take lots of time and planning. I think it is best that we just dont use tags right now.

sorry for the late response, but what do you mean with this?
seems to me that if we implement the few tweaks you suggested, this is good to go

woodsaj:
> sorry for the late response, but what do you mean with this?
seems to me that if we implement the few tweaks you suggested, this is good to go

Dont use tags.  The metrictank stats library doesnt support it.  How we update the stats library to support tags requires thought and consideration.  That hasnt happened yet.  So 'we'll merge them back once we validate the approach' isnt ever going to happen.  We are just going to have 2 different stats libraries.

Also, it isn't easy to mix tagged and non-tagged series in the same query.   If we are going to use tags, we should use them everywhere.

Dieterbe:
PTAL

Dieterbe:
maybe. not sure. no rush to do that anyway

PR #172, opened 2019-07-05
Support publishing to multiple Kafka topics with restriction by org id

fkaleo:
Support publishing to multiple Kafka topics with restriction by org id:
- in the Kafka publisher: options "metrics-topic" and "metrics-partition-scheme" were amended to support multiple comma separated values; they remain backwards compatible
- in the Kafka publisher: a new option "only-org-id" was added which restricts which data get sent to each Kafka topic depending on the data's org id. An empty value or 0 means unrestricted
- a completion of vendored sarama was necessary to access the mocks for testing
- Int64SliceFlag was moved to the util package to allow reuse

### DIFF
--- a/util/flags.go
+++ b/util/flags.go
@@ -37,6 +37,6 @@ func (i *Int64SliceFlag) Set(value string) error {
 }
 
 func (i *Int64SliceFlag) String() string {
-	// This is just a 1-liner to convert print a slice as a command separated list.
+	// This is just a 1-liner to print a slice as a comma separated list.
 	return strings.Trim(strings.Replace(fmt.Sprint(*i), " ", ", ", -1), "[]")
 }
--- a/util/flags.go
+++ b/util/flags.go
@@ -2,6 +2,9 @@ package util
 
 import (
 	"flag"
+	"fmt"
+	"strconv"
+	"strings"
 )
 
 // Registerer is a thing that can RegisterFlags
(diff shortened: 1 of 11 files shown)


util/flags_test.go
```diff
@@ -0,0 +1,64 @@
...
+	})
+
+	Convey("When setting Int64SliceFlag to many orgs", t, func(c C) {
+		intSliceFlag := Int64SliceFlag{}
+		err := intSliceFlag.Set("10,1,17")
+		c.So(err, ShouldBeNil)
+		c.So(intSliceFlag, ShouldHaveLength, 3)
+		c.So(intSliceFlag[0], ShouldEqual, 10)
+		c.So(intSliceFlag[1], ShouldEqual, 1)
+		c.So(intSliceFlag[2], ShouldEqual, 17)
```
Dieterbe: at this point, don't worry about it, but in the future, let's not use convey and just use https://github.com/google/go-cmp to compare against `[]int64{10,1,1174}`

Dieterbe:
should we add validation checks to make sure metrics-topic, metrics-partition-scheme and onlyOrgIds have the same length (when length is > 1)

util/flags_test.go
fkaleo: noted

fkaleo:
> should we add validation checks to make sure metrics-topic, metrics-partition-scheme and onlyOrgIds have the same length (when length is > 1)

Fixed in latest commit

util/flags.go
```diff
@@ -15,3 +18,25 @@ func RegisterFlags(rs ...Registerer) {
...
+		if err != nil {
+			return err
+		}
+		*i = append(*i, int64(parsed))
+	}
+	return nil
+}
+
+func (i *Int64SliceFlag) String() string {
+	// This is just a 1-liner to convert print a slice as a command separated list.
```
robert-milan:
```suggestion
	// This is just a 1-liner to print a slice as a comma separated list.
```

PR #175, opened 2019-08-06
 Added option 'discard-prefixes' that discards data points of series starting with the given prefixes

fkaleo:
Added option 'discard-prefixes' that discards data points of series starting with the given prefixes.

This is needed during instance migrations so that the target instance does not receive metrics that are internal/specific to the source instance such as:
`tsdb-gw.stats.local-cluster-10-cassandra.tsdb-10-byseries-32-cassandra-55fdd66b98-86xwq`
`metrictank.stats.local-cluster-10-cassandra.mt-read00-10-byseries-32-cassandra-a-758966f4c8-8lk76`

### DIFF
--- a/publish/kafka/publish.go
+++ b/publish/kafka/publish.go
@@ -36,6 +36,7 @@ var (
 
 	topicsStr           string
 	onlyOrgIds          util.Int64SliceFlag
+	discardPrefixesStr  string
 	codec               string
 	enabled             bool
 	partitionSchemesStr string
@@ -50,9 +51,10 @@ var (
 )
 
 type topicSettings struct {
-	name        string
-	partitioner *p.Kafka
-	onlyOrgId   int
+	name            string
+	partitioner     *p.Kafka
+	onlyOrgId       int
+	discardPrefixes []string
 }
 
 type mtPublisher struct {
@@ -68,6 +70,7 @@ type Partitioner interface {
 func init() {
 	flag.StringVar(&topicsStr, "metrics-topic", "mdm", "topic for metrics (may be given multiple times as a comma-separated list)")
 	flag.Var(&onlyOrgIds, "only-org-id", "restrict publishing data belonging to org id; 0 means no restriction (may be given multiple times, once per topic, as a comma-separated list)")
+	flag.StringVar(&discardPrefixesStr, "discard-prefixes", "", "discard data points starting with one of the given prefixes separated by | (may be given multiple times, once per topic, as a comma-separated list)")
 	flag.StringVar(&codec, "metrics-kafka-comp", "snappy", "compression: none|gzip|snappy")
 	flag.BoolVar(&enabled, "metrics-publish", false, "enable metric publishing")
 	flag.StringVar(&partitionSchemesStr, "metrics-partition-scheme", "bySeries", "method used for paritioning metrics. (byOrg|bySeries) (may be given multiple times, once per topic, as a comma-separated list)")
@@ -94,15 +97,22 @@ func getCompression(codec string) sarama.CompressionCodec {
 	}
 }
 
-func parseTopicSettings(partitionSchemesStr, topicsStr string, onlyOrgIds []int64) ([]topicSettings, error) {
+func parseTopicSettings(partitionSchemesStr, topicsStr string, onlyOrgIds []int64, discardPrefixesStr string) ([]topicSettings, error) {
 	var topics []topicSettings
 	partitionSchemes := strings.Split(partitionSchemesStr, ",")
 	topicsStrList := strings.Split(topicsStr, ",")
+	var discardPrefixesStrList []string
+	if discardPrefixesStr != "" {
+		discardPrefixesStrList = strings.Split(discardPrefixesStr, ",")
+	}
 	if len(partitionSchemes) > 1 && len(partitionSchemes) != len(topicsStrList) {
 		return nil, errors.New("More partition schemes (metrics-partition-scheme) than topics (metrics-topic)")
 	}
 	if len(onlyOrgIds) > 1 && len(onlyOrgIds) != len(topicsStrList) {
 		return nil, errors.New("More org ids (only-org-id) than topics (metrics-topic)")
+	}
+	if len(discardPrefixesStrList) > 0 && len(discardPrefixesStrList) != len(topicsStrList) {
+		return nil, errors.New("More discard prefixes (discard-prefixes) than topics (metrics-topic)")
 	}
 	for i, topicName := range topicsStrList {
 		topicName = strings.TrimSpace(topicName)
@@ -127,10 +137,19 @@ func parseTopicSettings(partitionSchemesStr, topicsStr string, onlyOrgIds []int6
 		} else {
 			onlyOrgId = onlyOrgIds[i]
 		}
+
+		var discardPrefixes []string
+		if len(discardPrefixesStrList) == 0 {
+			discardPrefixes = nil
+		} else {
+			discardPrefixes = strings.Split(discardPrefixesStrList[i], "|")
+		}
+
 		topic := topicSettings{
-			name:        topicName,
-			partitioner: partitioner,
-			onlyOrgId:   int(onlyOrgId),
+			name:            topicName,
+			partitioner:     partitioner,
+			onlyOrgId:       int(onlyOrgId),
+			discardPrefixes: discardPrefixes,
 		}
 		topics = append(topics, topic)
 	}
@@ -151,7 +170,7 @@ func New(broker string, autoInterval bool) *mtPublisher {
 		autoInterval: autoInterval,
 	}
 
-	mp.topics, err = parseTopicSettings(partitionSchemesStr, topicsStr, onlyOrgIds)
+	mp.topics, err = parseTopicSettings(partitionSchemesStr, topicsStr, onlyOrgIds, discardPrefixesStr)
 	if err != nil {
 		log.Fatalf("failed to initialize partitioner: %s", err)
 	}
--- a/publish/kafka/publish.go
+++ b/publish/kafka/publish.go
@@ -306,7 +306,16 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
 				return err
 			}
 
-			if topic.onlyOrgId == 0 || metric.OrgId == topic.onlyOrgId {
+			prefixDiscarded := false
+			for _, prefix := range topic.discardPrefixes {
+				if strings.HasPrefix(metric.Name, prefix) {
+					prefixDiscarded = true
+					break
+				}
+			}
+
+			if (topic.onlyOrgId == 0 || metric.OrgId == topic.onlyOrgId) &&
+				!prefixDiscarded {
 				message := &sarama.ProducerMessage{
 					Key:   sarama.ByteEncoder(key),
 					Topic: topic.name,
--- a/publish/kafka/publish.go
+++ b/publish/kafka/publish.go
@@ -138,23 +138,20 @@ func parseTopicSettings(partitionSchemesStr, topicsStr string, onlyOrgIds []int6
 			onlyOrgId = onlyOrgIds[i]
 		}
 
-		var discardPrefixes []string
-		if len(discardPrefixesStrList) == 0 {
-			discardPrefixes = nil
-		} else {
+		topic := topicSettings{
+			name:        topicName,
+			partitioner: partitioner,
+			onlyOrgId:   int(onlyOrgId),
+		}
+
+		if len(discardPrefixesStrList) > 0 {
 			// split prefixes by '|'; similar to strings.Split() but removes empty splits
 			f := func(c rune) bool {
 				return c == '|'
 			}
-			discardPrefixes = strings.FieldsFunc(discardPrefixesStrList[i], f)
+			topic.discardPrefixes = strings.FieldsFunc(discardPrefixesStrList[i], f)
 		}
 
-		topic := topicSettings{
-			name:            topicName,
-			partitioner:     partitioner,
-			onlyOrgId:       int(onlyOrgId),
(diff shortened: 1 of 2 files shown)


Dieterbe:
as always, a PR should describe why we want this.

fkaleo:
> as always, a PR should describe why we want this.

sorry, edited the top comment to add that

Dieterbe:
we do not send per-pod stats into an instance.
we only send the aggstats into an instance,
these metrics match the pattern `^(metrictank|tsdb-gw)\.aggstats\.<instance-name>`
where currently `<instance-name>` is like `clustername-id-name` e.g. `ops-us-east-12574-ops`

at the time of writing, these are generated as aggregates in carbon-relay-ng like so:
```
   'addAgg avg ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.latency\.(median|mean|p75)\.gauge32 tsdb-gw.aggstats.$1.api.request.$2.latency.$3.gauge32 10 22',
    'addAgg max ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.latency\.(max|p90)\.gauge32 tsdb-gw.aggstats.$1.api.request.$2.latency.$3.gauge32 10 22',
    'addAgg min ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.latency\.min\.gauge32 tsdb-gw.aggstats.$1.api.request.$2.latency.min.gauge32 10 22',
    'addAgg avg ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.size\.(median|mean|p75)\.gauge32 tsdb-gw.aggstats.$1.api.request.$2.size.$3.gauge32 10 22',
    'addAgg max ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.size\.(max|p90)\.gauge32 tsdb-gw.aggstats.$1.api.request.$2.size.$3.gauge32 10 22',
    'addAgg min ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.size\.min\.gauge32 tsdb-gw.aggstats.$1.api.request.$2.size.min.gauge32 10 22',
    'addAgg sum ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.status\.([^\.]+)\.counter32 tsdb-gw.aggstats.$1.api.request.$2.status.$3.counter32 10 22',
    'addAgg sum ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.api\.request\.(metrics_find|render|metrics|metrics_index_json)\.status\.([^\.]+)\.rate32 tsdb-gw.aggstats.$1.api.request.$2.status.$3.rate32 10 22',
    'addAgg sum ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.(.*counter32) tsdb-gw.aggstats.$1.metrics.$2 10 22',
    'addAgg sum ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.(.*rate32) tsdb-gw.aggstats.$1.metrics.$2 10 22',
    'addAgg sum ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.output\.kafka\.published\..*\.counter32 tsdb-gw.aggstats.$1.metrics.published.counter32 10 22',
    'addAgg sum ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.output\.kafka\.published\..*\.rate32 tsdb-gw.aggstats.$1.metrics.published.rate32 10 22',
    'addAgg avg ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.publish\.latency\.(median|mean|p75)\.gauge32 tsdb-gw.aggstats.$1.metrics.publish.latency.$2.gauge32 10 22',
    'addAgg max ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.publish\.latency\.(max|p90)\.gauge32 tsdb-gw.aggstats.$1.metrics.publish.latency.$2.gauge32 10 22',
    'addAgg min ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.publish\.latency\.min\.gauge32 tsdb-gw.aggstats.$1.metrics.publish.latency.min.gauge32 10 22',
    'addAgg avg ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.message_size\.(median|mean|p75)\.gauge32 tsdb-gw.aggstats.$1.metrics.message_size.$2.gauge32 10 22',
    'addAgg max ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.message_size\.(max|p90)\.gauge32 tsdb-gw.aggstats.$1.metrics.message_size.$2.gauge32 10 22',
    'addAgg min ^tsdb-gw\.stats\.([^\.]+)\.[^\.]+\.metrics\.message_size\.min\.gauge32 tsdb-gw.aggstats.$1.metrics.message_size.min.gauge32 10 22',
    'addAgg max ^metrictank\.stats\.([^\.]+)\.mt-read.*\.api\.request\.render\.points_fetched\.(max|p90)\.gauge32 metrictank.aggstats.$1.api.request.render.points_fetched.$2.gauge32 10 22',
    'addAgg avg ^metrictank\.stats\.([^\.]+)\.mt-read.*\.api\.request\.render\.points_fetched\.median\.gauge32 metrictank.aggstats.$1.api.request.render.points_fetched.median.gauge32 10 22',
    'addAgg max ^metrictank\.stats\.([^\.]+)\.mt-read.*\.api\.request\.render\.series\.(max|p90)\.gauge32 metrictank.aggstats.$1.api.request.render.series.$2.gauge32 10 22',
    'addAgg avg ^metrictank\.stats\.([^\.]+)\.mt-read.*\.api\.request\.render\.series\.median\.gauge32 metrictank.aggstats.$1.api.request.render.series.median.gauge32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.tank\.metrics_too_old\.counter32 metrictank.aggstats.$1.tank.metrics_too_old.counter32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.tank\.metrics_too_old\.rate32 metrictank.aggstats.$1.tank.metrics_too_old.rate32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.tank\.add_to_closed_chunk\.counter32 metrictank.aggstats.$1.tank.add_to_closed_chunk.counter32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.tank\.add_to_closed_chunk\.rate32 metrictank.aggstats.$1.tank.add_to_closed_chunk.rate32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.input\..*\.(metric_invalid|metrics_decode_err)\.counter32 metrictank.aggstats.$1.input.$2.counter32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.input\.[^\.]+\.[^\.]+\.invalid.counter32 metrictank.aggstats.$1.input.metric_invalid.counter32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.input\.[^\.]+\.metrics_decode_err\.rate32 metrictank.aggstats.$1.input.metrics_decode_err.rate32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.input\.[^\.]+\.[^\.]+\.invalid.rate32 metrictank.aggstats.$1.input.metric_invalid.rate32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.idx\.metrics_active\.gauge32 metrictank.aggstats.$1.idx.metrics_active.gauge32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-write.*\.tank\.metrics_active\.gauge32 metrictank.aggstats.$1.tank.metrics_active.gauge32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-read.*\.cache\.ops\.metric\.(hit.*|miss)\.counter32 metrictank.aggstats.$1.cache.ops.metric.$2.counter32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-read.*\.cache\.ops\.metric\.(hit.*|miss)\.rate32 metrictank.aggstats.$1.cache.ops.metric.$2.rate32 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-read.*\.cache\.size\.max\.gauge64 metrictank.aggstats.$1.cache.size.max.gauge64 10 22',
    'addAgg sum ^metrictank\.stats\.([^\.]+)\.mt-read.*\.cache\.size\.used\.gauge64 metrictank.aggstats.$1.cache.size.used.gauge64 10 22',
```

We have 3 choices:
### 1) keeping the naming scheme as is, such that a migration creates a new instance that exposes metrics with the same name as before

what would happen during a migration is that stats from the old and the new instance would have the same name and timestamp and thus be dropped (but increment the dropped counter which could confuse our customers!). so in that case I agree each instance should only ingest the stats of one of the instances (either those of itself or the other one).
as far as which should be dropped:
* both of the old and new instance should ingest the same data, so those stats should be equivalent, so it wouldn't matter which we drop.
* the most useful stats wrt queries/reads is that of whichever instance is being queried, but i'm not seeing a good way to persist that metrics stream of whichever instance is receiving queries, not to mention that the query workload may switch over gradually. so i think we'd just have to take the hit and live with the fact that the stats may not be accurate during a migration.

but then the proposed solution wouldn't work here: you can't match the metrics to drop because the new and old instance metrics would have the same name.

instead what we could do is:
* make sure only 1 carbon-relay-ng is alive at a time
* manage the carbon-relay-ng config such that only one is sending the aggregates to their tsdbgw service.
* adjust the multi-topic publishing feature such that the aggstats only get sent to the primary kafka topic and not the secondary ("other instance")

### 2) transitioning to a naming scheme with "richer" instance names that include partitioning scheme, num partitions, etc. such that any migration would result in a distinct name and a distinct set of metrics that wouldn't conflict with one another.

* exposes more of the instance configuration details to the customer
* depending on users have their queries/dashboards/alerts set up, it may require them to adjust their configuration (if it assumes a specific name), or may double count some metrics (if they have a wildcard for instance name)

### 3) transitioning to a naming scheme that does not include the instance name at all
this would itself already be a breaking change and break users dashboards/queries/alerts, so i'm very hesitant to do this.. but assuming we get past that, then the rest of the story is as in case 1.

Dieterbe:
at first glance, my preferred choice is the 3rd solution under point 1.

fkaleo:
> at first glance, my preferred choice is the 3rd solution under point 1.
> it seems that is supported by this PR. i'm not sure if this PR was intended to be used like this. to be clear, the first topic would not have any prefix to discard, only the 2nd one would drop everything matching metrictank.aggstats and tsdb-gw.aggstats

Took me a while to ingest your comments, sorry :) Essentially I agree with your preferred choice. This PR is made to support that indeed. In the migration PR we set `GW_DISCARD_PREFIXES` to something like `,metrictank.aggstats|tsdb-gw.aggstats` so that the first topic, the one we published to originally, stays unchanged and receives everything, but the second topic, the additional topic, receives everything minus the aggregated statistics. Does that mean we are good here?


Of course, we still have work/discussion to do for what you said on Slack but not in this PR I think?

> 1. ingestion rate. (exposed by tsdb-gw, in the publisher)
> 2. active series (exposed by MT)
> 3a. dropped samples (exposed by tsdb-gw)
> 3b. dropped samples (exposed by MT)
> for 1 and 3a, we should always just add up across the old/new instance and the totals will always make sense regardless of where we are in the migration process (i.o.w. regardless of which tsdbgw/instance publishers are writing to)
> for 2 and 3b we should ideally only count across one instance, to avoid double-counting. i imagine we probably currently are already not very accurate here: i assume we only count across write nodes (to have non-duplicated numbers due to replication) but even then, when we restart pods, it would mess up the numbers. @awoods or @Jacob Lisi do you know if we can make the billing scrapers "smart" and point them to only 1 instance at a time and switch them over when we're ready? cc @Kaleo (Florian)

Dieterbe:
>  In the migration PR we set GW_DISCARD_PREFIXES to something like ,metrictank.aggstats|tsdb-gw.aggstats so that the first topic, the one we published to originally, stays unchanged and receives everything, but the second topic, the additional topic, receives everything minus the aggregated statistics. Does that mean we are good here?

sounds good. i expect to approve this pr shortly after i look at it a bit more.

Dieterbe:
the counts of published metrics messages, and the debug log message is no longer correct
https://github.com/raintank/tsdb-gw/pull/175/files#diff-c1de7345bb1a47b5638ddad84e8ae8eeR356

due to: filtering by prefix and org id, and due to multi-topic
perhaps this should be fixed in a separate PR though.
note: i would rather not modify our metric naming scheme again (e.g. include the topic in the key to have multi-topic stats), so we can just keep using the same metrics

fkaleo:
> the counts of published metrics messages, and the debug log message is no longer correct
> https://github.com/raintank/tsdb-gw/pull/175/files#diff-c1de7345bb1a47b5638ddad84e8ae8eeR356
> 
> due to: filtering by prefix and org id, and due to multi-topic
> perhaps this should be fixed in a separate PR though.
> note: i would rather not modify our metric naming scheme again (e.g. include the topic in the key to have multi-topic stats), so we can just keep using the same metrics

I made a tentative fix in #176. Let me know what you think.

PR #177, opened 2019-08-13
Synchronised qa/ and qa-subjective/ scripts with metrictank and applied same CI rules

fkaleo:


api/api.go
```diff
@@ -67,8 +67,8 @@ func (a *Api) Start() *Api {
 		defer close(a.done)
 		var err error
 		if *ssl {
-			cert, err := tls.LoadX509KeyPair(*certFile, *keyFile)
-			if err != nil {
+			cert, errKeyPair := tls.LoadX509KeyPair(*certFile, *keyFile)
+			if errKeyPair != nil {
 				log.Fatalf("Fail to start server: %v", err)
```
Dieterbe:
```suggestion
				log.Fatalf("Fail to start server: %v", errKeyPair)
```

scripts/qa/gofmt.sh
```diff
@@ -0,0 +1,18 @@
...
+# find the dir we exist within...
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+# and cd into root project dir
+cd ${DIR}/../..
+
+out=$(gofmt -d -s $(find . -name '*.go' | grep -v vendor | grep -v _gen.go))
+if [ "$out" != "" ]; then
+	echo "$out"
+	echo
+	echo "You might want to run something like 'find . -name '*.go' | xargs gofmt -w -s'"
```
robert-milan:
```suggestion
	echo "You might want to run something like 'find . -name '*.go' -not -path './vendor/*' | xargs gofmt -w -s'"
```

I know it's not hard to revert the changes to vendor in git, but this is easier. We should change it to this everywhere else in our code as well.

fkaleo:
> Other than that and the one suggestion to the `find` suggestion

I updated Metrictank to do the same: https://github.com/grafana/metrictank/pull/1437

fkaleo:
> I think we should look into using https://github.com/golangci/golangci-lint and replace anything we can with it. I ended up there by following links from `misspell`

Seems pretty nice

fkaleo:
> [...] but I think we should add options to the `Makefile` to use all of these scripts locally as well, just like in `Metrictank`.

Done in commit e673ce5 : new 'make qa'

fkaleo:
> A few more changes are required in the Makefile. Specifically `deps:` will now fail. We should also remove all of the `bash -c ...` unless there is some reason we still need it. Also let's add a new line at the end of the Makefile for good measure.
> 
> Also, do we want to use the `-race` flag here for `go test ...`?

Last 4 commits should address all these, thanks!

Issue #134, opened 2018-11-02
ingestion Rate Limits


DanCech:
Currently if a customer starts to send more metrics than their deployment is sized for we can run into issues with tsdb-gw or other components in the stack using too much memory and being killed by k8s, etc.  This results in the user getting nasty error messages like 504 gateway timeouts from nginx or 500 backend errors.

It would be much nicer if tsdb-gw had a rate-limiting capability that we could set to protect the system and instead return a 429 error that explains to the user that they're exceeding their provisioned capacity and should contact us to adjust their plan.

This would also be helpful in the multi-tenant deployments as it would give us a way to prevent a single user from vastly exceeding the capacity we expect them to use on the cluster.

Dieterbe:
this means we should also do experiments and confirm what amounts of rate/s, active series, etc correspond to which amounts of ram and cpu.

replay:
how does that interact with POSTs that push a large number of metrics in one request? f.e. if a user has a ratelimit of 1000 datapoints/sec, then they submit a single POST which includes 1001 datapoints, should the response be a 429 after 1000 have been processed? In that situation CRNG will resend all the 1001 datapoints, while holding off with newer datapoints to prevent out of order, resulting in a loop which blocks everything. So we should probably only reply with 429 to the requests where all included metrics are above the limit and process the previous requests fully even if we exceed the limit by doing that.

tomwilkie:
Not sure this issue is relevant for cortex-gw (which lives in this repo), as Cortex implements all these rate limits itself.  I'd like to ensure cortex-gw remains stateless (therefore it can't track active series) and doesn't parse the requests (which can be expensive and I don't want to add any more latency).

Dieterbe:
seeing this affect another HM customer that was increasing their MT index size and resulting in slow queries.
I think we should bump the priority of this

Dieterbe:
let's keep this ticket for rate limit on datapoint volume.
for other limits, such as on active series, request workload etc, we have another project in the process of being specced out.

PR #182, opened 2019-08-21
add optional rate limiter

replay:
Related #134

I'm not sure what to do if a single request contains more datapoints than the secondly budget. 
I'd assume that this is a case that shouldn't happen unless we limit a user to an extremely low rate, or a user configures their crng to a huge batch size.
I think we could enforce something like a max batch size by responding with 429 if the received batch has more datapoints than the configured secondly rate. If a user does receive 429s they'd then have to configure their crng's max-batch-size.

### DIFF
--- a/cmd/tsdb-gw/main.go
+++ b/cmd/tsdb-gw/main.go
@@ -54,6 +54,7 @@ var (
 
 	// limitations
 	timerangeLimit = flag.String("timerange-limit", "", "define maximum timerange to serve queries for")
+	rateLimits     = flag.String("rate-limits", "", "define rate limits in the format \"<orgId>:<limit>;<orgId>;<limit>\"")
 
 	metricsAddr = flag.String("metrics-addr", ":8001", "http service address for the /metrics endpoint")
 )
@@ -119,6 +120,10 @@ func main() {
 		if err := ingest.InitMtBulkImporter(*importerURL); err != nil {
 			log.Fatalf(err.Error())
 		}
+	}
+
+	if err := ingest.ConfigureRateLimits(*rateLimits); err != nil {
+		log.Fatalf(err.Error())
 	}
 
 	inputs := make([]Stoppable, 0)
--- a/cmd/tsdb-gw/main.go
+++ b/cmd/tsdb-gw/main.go
@@ -54,7 +54,7 @@ var (
 
 	// limitations
 	timerangeLimit = flag.String("timerange-limit", "", "define maximum timerange to serve queries for")
-	rateLimits     = flag.String("rate-limits", "", "define rate limits in the format \"<orgId>:<limit>;<orgId>:<limit>\"")
+	rateLimits     = flag.String("rate-limits", "", "define rate limits in the format \"<orgId>:<limit>;<orgId>:<limit>\" where <limit> is the number of datapoints per second")
 
 	metricsAddr = flag.String("metrics-addr", ":8001", "http service address for the /metrics endpoint")
 )
--- a/cmd/tsdb-gw/main.go
+++ b/cmd/tsdb-gw/main.go
@@ -54,7 +54,7 @@ var (
 
 	// limitations
 	timerangeLimit = flag.String("timerange-limit", "", "define maximum timerange to serve queries for")
-	rateLimits     = flag.String("rate-limits", "", "define rate limits in the format \"<orgId>:<limit>;<orgId>;<limit>\"")
+	rateLimits     = flag.String("rate-limits", "", "define rate limits in the format \"<orgId>:<limit>;<orgId>:<limit>\"")
 
 	metricsAddr = flag.String("metrics-addr", ":8001", "http service address for the /metrics endpoint")
 )
--- a/ingest/metrics.go
+++ b/ingest/metrics.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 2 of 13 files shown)


cmd/tsdb-gw/main.go
```diff
@@ -54,6 +54,7 @@ var (
 
 	// limitations
 	timerangeLimit = flag.String("timerange-limit", "", "define maximum timerange to serve queries for")
+	rateLimits     = flag.String("rate-limits", "", "define rate limits in the format \"<orgId>:<limit>;<orgId>:<limit>\"")
```
robert-milan: What exactly is the format of `<limit>`? I guess we just add this to our documentation efforts.

ingest/metrics.go
```diff
@@ -34,8 +37,54 @@ var (
...
+		if err != nil {
+			return fmt.Errorf("Unable to parse orgId from string: %q", limitWithOrgParts[0])
+		}
+
+		limit, err := strconv.ParseInt(limitWithOrgParts[1], 10, 32)
+		if err != nil {
+			return fmt.Errorf("Unable to parse rate limit from string: %q", limitWithOrgParts[1])
+		}
+
+		rateLimiters[int(orgId)] = rate.NewLimiter(rate.Limit(limit), int(limit))
```
robert-milan: We could increase the burst size here, which would partially address your comment. I don't know that is the best solution. What about creating a new metric to track how often batch sizes are greater than the rate?

woodsaj:
i have thought about this some more, and i dont think this approach will prevent OOM issues.

The rate limiting is being applied after the metrics payloads have been decoded, and the limiting approach is to just wait until the limit is not being exceeded.  

So once the limit is reached, new connections will be accepted and memory will be allocated to decode the payloads, then the requests will just get blocked. This will certainly lead to OOM issues.


I think a better approach is to immediately reject connections if the rate limits have been reached.

Additionally it might be simpler to just limit concurrent connections.

ingest/metrics.go
woodsaj: I dont think we will run into problems with batch sizes being too large.  the default flushMaxNum in crng is 5000.
But we could add an additional check and return a different error message if the batch is larger than the limit, so it will be more obvious to users that they have misconfigured their crng instance.

replay:
My assumption was that all carbon-relay-ng instances are using a static number of connections (their concurrency) when posting into tsdb. F.e. if a crng runs at concurrency of 10, and it hits the rate limit, then all 10 connections simply get blocked until the end of the second because we delay their post response. I assumed holding the decoded (batch size * number connections) datapoints in memory should be fine. but i think it would be good to limit the batch size to something reasonable, and reply with a 429 if the batch size is too big, that way we can set a limit on how much memory each connection can consume.

replay:
if we immediately reject connections once the rate limit has been reached, then we can only limit the number of requests per time, but we don't know how many datapoints each request contains. i guess that works too, then we just need to make sure that the sending crngs don't get configured to a huge batch size, otherwise they could still over load the tsdb.

replay:
We could also try some mixed approach:

When a new connection comes in we check in the rate limiter if the secondly budget has already been exhausted or not. If yes, we directly reject the request without decoding. If no, then we decode the request and handle it, and subtract the number of datapoints from the secondly budget. 
The secondly budget would still be defined in number of datapoints per second, not number of connections per second. That way we'd often still handle a bit more datapoints than the rate limit is allowing, f.e. if the secondly budget has 500 datapoints left and a request with 1000 datapoints comes in, then we'd still handle all of them. But after that, for the rest of the second, subsequent requests would be rejected.

What do you think about that @awoods ^^

cmd/tsdb-gw/main.go
replay: it's just the number of datapoints per second, as a number, i'll add that to the description

replay:
I think the latest version that I pushed should be a good combination of the different approaches. It still limits the number of data points per second, once they are decoded it checks the rate limit and waits before publishing if necessary, but additionally it rejects all new requests with 429 as soon as the secondly budget is exhausted. 
That way, if a crng is trying to publish at a rate that's above the limit, we'll always just hold a small number of decoded request in memory (per org) which are waiting until they are allowed to go ahead while all others are getting a 429 until there is budget again.

Dieterbe:
> I assumed holding the decoded (batch size * number connections) datapoints in memory should be fine.

this would be equivalent to having no rate limiting (because even without rate limiting, we only respond to the request after the data has been flushed - thus we block the crng routine until the memory is released). having this much data in RAM is what caused the outage, thus it is not fine.

The revised mixed approach does make sense though.

ingest/rate_limit_test.go
```diff
@@ -0,0 +1,180 @@
...
+					if limiter.Limit() != rate.Limit(expectedLimit) {
+						t.Fatalf("Expected org %d to have limit %d, but it had %d", orgId, expectedLimit, int(limiter.Limit()))
+					}
+				}
+			}
+		})
+	}
+}
+
+// TestLimitingRate is a bit racy, but since it tests a rate limiter I can't think of a better way to do it
```
Dieterbe: it would really help if you gave the reader a bit more of a clue about what you're talking about here?
should we just add "see below comments" or are you referring to our inability to accurately sync the timing of the limiter and the requests against it?

Dieterbe: @replay ^

PR #191, opened 2019-10-07
groupByNode average, not avg

Dieterbe:
per
https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.groupByNodes
'avg' is not supported and leads to http 400's when
GRAPHITE_ENFORCE_INPUT_VALIDATION is enabled

see also https://github.com/graphite-project/graphite-web/issues/2494


Dieterbe:
actually looks like a problem with the graphite validation that'll be fixed

Dieterbe:
the validation is now patched

PR #193, opened 2019-10-15
Allow multi broker

Dieterbe:


Dieterbe:
no. dockerhub is down right now

cmd/tsdb-gw/main.go
```diff
@@ -99,7 +99,7 @@ func main() {
 	}
 	defer traceCloser.Close()
 
-	publisher := kafka.New(*broker, true)
+	publisher := kafka.New(strings.Split(*brokers, ","), true)
```
robert-milan: We might want to add some error checking on the split string as it could prove difficult to debug further down in the code if problems arise. (i.e. the first one is fine but the others aren't, so it always connects to that one and you don't immediately see problems, or something similar).

PR #196, opened 2019-10-28
Remove cortex-gw and update promstats to `tsdb_gw_*`

jtlisi:
Signed-off-by: Jacob Lisi <jacob.t.lisi@gmail.com>


Dieterbe:
there's still a couple references to "cortex" in the code base. can you clean it up? also scripts/config/cortex-gw.ini

Dieterbe:
what's this
```
auth/gcom/auth_test.go
557:	validInstanceType = "cortex"
603:		InstanceType: "cortex",
606:	validInstanceType = "cortex"
```

jtlisi:
@Dieterbe that's test cases for auth which is tied into the `/api/hosted-metrics` endpoint on Grafana.com so that test case would still be valid for that package.

PR #198, opened 2020-01-27
Add metrictank proxy endpoint for /tags/delSeries

fitzoh:
Part of https://github.com/grafana/metrictank-ops/issues/554, proxies to 

https://github.com/grafana/metrictank/blob/01a5b0536902f3c48aee0d740fbe5a6ac6b7b719/api/routes.go#L72


fitzoh:
So the Metrictank endpoint has a `propagate` param which ensures that the request is sent to all cluster nodes.... I'm thinking we want to modify our proxy to set `propagate=true`?

replay:
> So the Metrictank endpoint has a propagate param which ensures that the request is sent to all cluster nodes.... I'm thinking we want to modify our proxy to set propagate=true?

Not necessary anymore after https://github.com/grafana/metrictank/pull/1637

PR #199, opened 2020-02-05
New flag "rewrite-org-id" allowing orgId to be rewritten when published to Kafka

### LINKED ISSUE #197: New feature: ability to rewrite the org id per topic
New feature:
ability to rewrite the org id per topic.
We should then be able to specify a list of kafka topics together with orgids that we want to rewrite the data to in the tsdb-gw config

Necessary for instance migrations.

fkaleo:
New flag "rewrite-org-id" allowing `orgId` to be rewritten when published to Kafka.
This will be useful in instance migrations where the orgId is already used.

Can be specified once for each Kafka topic.
For example:
`--rewrite-org-id=33:45,1:10` will rewrite `orgId` for all incoming `MetricData` with `orgId` `33` to `45`, and the ones with `orgId` `1` to `10`. 

Some light refactoring was necessary, in particular isolating the marshaling of `MetricData`.

Fixes #197

### DIFF
--- a/publish/kafka/publish.go
+++ b/publish/kafka/publish.go
@@ -35,6 +35,7 @@ var (
 	sendErrOther    = stats.NewCounterRate32("metrics.send_error.other")
 
 	topicsStr           string
+	rewriteOrgIdStr     string
 	onlyOrgIds          util.Int64SliceFlag
 	discardPrefixesStr  string
 	codec               string
@@ -56,6 +57,10 @@ type topicSettings struct {
 	numPartitions   int32
 	onlyOrgId       int
 	discardPrefixes []string
+	orgIdRewrite    struct {
+		source int
+		target int
+	}
 }
 
 type mtPublisher struct {
@@ -70,6 +75,7 @@ type Partitioner interface {
 
 func init() {
 	flag.StringVar(&topicsStr, "metrics-topic", "mdm", "topic for metrics (may be given multiple times as a comma-separated list)")
+	flag.StringVar(&rewriteOrgIdStr, "rewrite-org-id", "", "rewrite org id; for example 33:45 means rewriting org id 33 into 45 (may be given multiple times, once per topic, as a comma-separated list)")
 	flag.Var(&onlyOrgIds, "only-org-id", "restrict publishing data belonging to org id; 0 means no restriction (may be given multiple times, once per topic, as a comma-separated list)")
 	flag.StringVar(&discardPrefixesStr, "discard-prefixes", "", "discard data points starting with one of the given prefixes separated by | (may be given multiple times, once per topic, as a comma-separated list)")
 	flag.StringVar(&codec, "metrics-kafka-comp", "snappy", "compression: none|gzip|snappy")
@@ -98,7 +104,7 @@ func getCompression(codec string) sarama.CompressionCodec {
 	}
 }
 
-func parseTopicSettings(partitionSchemesStr, topicsStr string, onlyOrgIds []int64, discardPrefixesStr string) ([]topicSettings, error) {
+func parseTopicSettings(partitionSchemesStr, topicsStr string, onlyOrgIds []int64, discardPrefixesStr string, rewriteOrgIdStr string) ([]topicSettings, error) {
 	var topics []topicSettings
 	partitionSchemes := strings.Split(partitionSchemesStr, ",")
 	topicsStrList := strings.Split(topicsStr, ",")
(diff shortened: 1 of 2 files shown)


publish/kafka/publish.go
```diff
@@ -115,6 +126,9 @@ func parseTopicSettings(partitionSchemesStr, topicsStr string, onlyOrgIds []int6
 	if len(discardPrefixesStrList) > 0 && len(discardPrefixesStrList) != len(topicsStrList) {
 		return nil, errors.New("More discard prefixes (discard-prefixes) than topics (metrics-topic)")
 	}
+	if len(rewriteOrgIdStrList) > 0 && len(rewriteOrgIdStrList) != len(topicsStrList) {
+		return nil, errors.New("More org id rewrites (rewrite-org-id) than topics (metrics-topic)")
```
replay: That error is confusing, it could also be less

publish/kafka/publish.go
```diff
@@ -70,6 +75,7 @@ type Partitioner interface {
 
 func init() {
 	flag.StringVar(&topicsStr, "metrics-topic", "mdm", "topic for metrics (may be given multiple times as a comma-separated list)")
+	flag.StringVar(&rewriteOrgIdStr, "rewrite-org-id", "", "rewrite org id; for example 33:45 means rewriting org id 33 into 45 (may be given multiple times, once per topic, as a comma-separated list)")
```
replay: do i understand it correctly that the number of rewrites must be the same as the number of topics? if so i think the help message could be a bit more explicit about it. 

```suggestion
	flag.StringVar(&rewriteOrgIdStr, "rewrite-org-id", "", "rewrite org id; for example 33:45 means rewriting org id 33 into 45 (may be given multiple times, if given then there must be exactly one per topic, as a comma-separated list)")
```

publish/kafka/publish.go
```diff
@@ -286,89 +371,72 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
...
-					_, err = mp.MarshalWithoutOrg28(data[:1])       // Marshal will fill up space between length and cap, i.e. bytes 2-29
-					isMPNO = true
+			}
+
+			if topic.onlyOrgId != 0 && metric.OrgId != topic.onlyOrgId {
+				continue TOPICS
+			}
+
+			orgID := metric.OrgId
+			if topic.orgIdRewrite.source != 0 {
```
replay: don't we have to check whether `topic.orgIdRewrite.source == metric.OrgId`?

publish/kafka/publish.go
```diff
@@ -286,89 +371,72 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
...
+				orgID = topic.orgIdRewrite.target
+			}
+
+			// retrieve MetricDataBuffer from cache if possible
+			mdBuffer, cached := mdBufferCache[orgID]
+			if !cached {
+				// rewrite orgId if needed
+				if topic.orgIdRewrite.source != 0 {
```
replay: i think this part is a bit redundant with `:391`. Why not just assume that the var `orgId` is the source of truth, so:

```
if metric.OrgId != orgId {
  metric.OrgId = orgId
  metric.SetId()
}
```

publish/kafka/publish.go
```diff
@@ -286,89 +371,72 @@ func (m *mtPublisher) Publish(metrics []*schema.MetricData) error {
...
-		}
 
-		buffersToRelease = append(buffersToRelease, data)
+				buffersToRelease = append(buffersToRelease, mdBuffer.Data)
+				mdBufferCache[orgID] = mdBuffer
+
+				// restore original orgId if needed
+				if topic.orgIdRewrite.source != 0 {
+					metric.OrgId = topic.orgIdRewrite.source
+					metric.SetId()
```
replay: If we set the id back to the original one here, won't the partitioner on `:418` potentially come to a different (wrong) result

publish/kafka/publish.go
fkaleo: Fixed

publish/kafka/publish.go
fkaleo: The idea was to avoid doing it (`SetId()` has a cost) if unnecessary

publish/kafka/publish.go
fkaleo: Fixed in 24ca805

publish/kafka/publish.go
fkaleo: Fixed in 24ca805

publish/kafka/publish.go
fkaleo: Fixed in 24ca805

replay: I don't see how this is fixed in https://github.com/raintank/tsdb-gw/commit/24ca8052a14417c7e738b22959d3fa7aebac5a25

fkaleo: You're right, I was deluded. fa9ead6 should do.

